PR #1, opened 2020-11-21
Ensure a dynlib extension is always returned.

dasmoth:


clauswilke:
Thanks, this is what I meant to write.

Issue #2, opened 2020-11-23
Building on Windows


CGMossa:
For Windows builds, the DLL is named without the "lib"-prefix. 

E.g. in `rust_source`:

```r

  # load shared library
  libfilename <- if (.Platform$OS.type == "windows") {
    paste0(libname, get_dynlib_ext())
  } else {
    paste0("lib", libname, get_dynlib_ext())
  }

  the$shared_lib <- file.path(dir, "target", "debug", libfilename)
  dyn.load(the$shared_lib, local = TRUE, now = TRUE)
```

### Common `CARGO_TARGET_DIR`

A common "trick" to reduce cargo's disk usage, and compile time, is to set [`CARGO_TARGET_DIR`](https://doc.rust-lang.org/cargo/reference/environment-variables.html) or other things, to one common place for all cargo packages. This means that, when `rextendr` tries to locate the `target`, it might not be able to do so properly, as those would be in the common target directory, and not in the project (which is in a temporary location) directory. 

Thus I've added:
```r
  status <- system2(
    command = "cargo",
    args = c(
      "build",
      #"--release",  # release vs debug should be configurable at some point; for now, debug compiles faster
      sprintf("--lib --manifest-path=%s --target-dir %s/target/", file.path(dir, "Cargo.toml"), dir)
    ),
    stdout = stdout,
    stderr = stdout
  )
```
to the `rust_source`.

### `dylib` vs. `cdylib`

While diagnosing my issues, I found this [thread](https://users.rust-lang.org/t/what-is-the-difference-between-dylib-and-cdylib/28847) describing the difference between `cdylib` and `dylib`. Turned out, it wasn't an issue for me to use `dylib`, but 
this might be a problem down the line.  Also, r-rust/hellorust uses `cdylib`.

### A note: The unlink command does not delete the "loaded" dll, as it is still in use. 
I tried adding a 

```r
clean_build_dir <- function() {
  if (!is.null(the$shared_lib))
    dyn.unload(x = the$shared_lib)
  if (!is.null(the$build_dir)) {
    unlink(the$build_dir, recursive = TRUE, force = TRUE)
    the$build_dir <- NULL
  }
}
```

But this is not good, as it is called after the build is done, and the the R-session is unable to call, since the dll is deleted.

I'll make a PR immediately as I get some comment on these points.

clauswilke:
[...]

PR #3, opened 2020-11-23
To build on Windows

CGMossa:
Change the lib-name without the prefix.
Specify the location of targets in the temporary folder.

Add "--lib" flag for insurance.

See #2 for discussion.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -4,7 +4,8 @@ Version: 0.0.0.9000
 Authors@R: c(
     person("Andy", "Thomason", , "andy@andythomason.com", c("aut", "cre")),
     person("Claus O.", "Wilke", , "wilke@austin.utexas.edu", "aut",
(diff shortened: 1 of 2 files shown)


clauswilke:
This looks good to me. Please add yourself to the authors in the DESCRIPTION file.

CGMossa:
I have added myself as author.

Please ensure, that I've added myself correctly, as I am new to this.

clauswilke:
Oh, sorry, I do have a few comments:

- The `args` argument is supposed to have a vector of individual arguments. Putting multiple ones into one string may break on some platforms. So please put one argument on each line. You'll then need two `sprintf()` calls I'm afraid.

- For the target dir, you'll also have to write `file.path(dir, "target")` to make sure the file path is put together correctly.

- As a general rule, it's a good idea to write something like "fixes #2" into one of the commit messages so github knows to close the issue when the PR is merged.

PR #7, opened 2020-11-23
Sourcing the exported extendr_wrappers.R

CGMossa:
This makes use of pull-request #40 on extendr/extendr.


clauswilke:
Doesn't this remove the need for all of this file: https://github.com/clauswilke/rextendr/blob/main/R/exported_functions.R

Also, with this PR we should make @hobofan a coauthor. His contribution is to remove 56 lines of code.

clauswilke:
And please don't add a test that cannot work because it looks for files only available on your hard drive. I had purposefully not added a test yet because it won't work until the various required crates are updated on crates.io. You can leave the code there but disable the test, or delete the test.

clauswilke:
@CGMossa Could you merge in the current main branch and check that this PR still works?

CGMossa:
Do you want me to press the squash and merge button or do you want me to pull from main and then push that pull onto this PR?

clauswilke:
I wanted you to do exactly what you just did. :-)

clauswilke:
Oh, but also try locally whether the resulting package still works. E.g., can you knit the vignette available here:
https://github.com/extendr/rextendr/blob/main/vignettes/rmarkdown.Rmd

CGMossa:
Alright, I'm having some issues:

```
error[E0463]: can't find crate for `std`                       
  |
  = note: the `x86_64-pc-windows-gnu` target may not be installed
```

This got fixed after I changed `rustup default stable-gnu`; It was not only missing 
target. It requires that the default rustup is not `msvc`, but `gnu`. I am mentioning this
as it was discussed that this will be passed to cargo directly.

Anyways, after this, the rmarkdown knits and everything works.

clauswilke:
Thanks!

PR #18, opened 2021-01-11
Add pkgdown files to .Rbuildignore

yutannihilation:
Fix this NOTE:

```
 * checking top-level files ... NOTE
Non-standard files/directories found at top level:
  '_pkgdown.yml' 'docs'
```

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -4,4 +4,5 @@
 ^README\.Rmd$
 ^vignettes/rmarkdown.Rmd$
 ^\.github$
-
+^docs^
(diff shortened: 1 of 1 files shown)


.Rbuildignore
```diff
@@ -4,4 +4,5 @@
 ^README\.Rmd$
 ^vignettes/rmarkdown.Rmd$
 ^\.github$
-
+^docs^
```
clauswilke, line 7: Should this end in `$`?

yutannihilation: Oh, of course... Thanks.

PR #20, opened 2021-01-12
Adding 'toolchain' parameter to control rust builds

### LINKED ISSUE #19: On Windows, default rust needs to be gnu, not msvc
Apparently, under some circumstances, the default Rust on Windows needs to be gnu, not msvc.

See:
https://github.com/extendr/rextendr/issues/15#issuecomment-757944529
and
https://github.com/extendr/rextendr/pull/7#issuecomment-758272287

Is there any way to remove this requirement?

cc @Ilia-Kosenkov @CGMossa @yutannihilation

Ilia-Kosenkov:
Fixes #19.
`toolchain` parameter allows overriding default toolchain used system-wide when building rust libraries.
In most cases, the default toolchain is the preferred choice, but on Windows default toolchain can be `*-msvc`, which requires `MSYS2` installed for linking. In such a scenario, `toolchain` allows using `*-gnu` to compile `rextendr` chunks without overriding system default.

### DIFF
--- a/R/source.R
+++ b/R/source.R
@@ -12,7 +12,8 @@
 #' @param profile Rust profile. Can be either `"dev"` or `"release"`. The default,
 #'   `"dev"`, compiles faster but produces slower code.
 #' @param toolchain Rust toolchain. The default, `NULL`, compiles with the
-#'  system default toolchain. Accepts valid Rust toolchain qualifiers.
+#'  system default toolchain. Accepts valid Rust toolchain qualifiers,
+#'  such as `"nightly"`, or (on Windows) `"stable-msvc"`.
 #' @param extendr_version Version of the extendr-api crate, provided as a Rust
 #'   version string. `"*"` will use the latest available version on crates.io.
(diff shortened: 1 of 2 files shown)


R/source.R
```diff
@@ -11,6 +11,8 @@
 #'   be added to the `Cargo.toml` file.
 #' @param profile Rust profile. Can be either `"dev"` or `"release"`. The default,
 #'   `"dev"`, compiles faster but produces slower code.
+#' @param toolchain Rust toolchain. The default, `NULL`, compiles with the
+#'  system default toolchain. Accepts valid Rust toolchain qualifiers.
```
clauswilke, line 15: I would add an example, just to be absolutely clear:

> Accepts valid Rust toolchain qualifiers, such as `"stable-msvc"` on Windows.

clauswilke: I just edited the above comment, in case you're seeing an old version in an email.

clauswilke:
Thanks!

PR #25, opened 2021-01-14
Add meaningful tests to rextendr

Ilia-Kosenkov:
In this PR I enable existing tests and add a few more -- to check boundary cases and wrapper generation.
Test are executed on *not CRAN* using `skip_on_cran()`.
This works out of the box on GHA because it sets `NOT_CRAN = "true"` by default.

I pass additional argument `--force-multiarch` to `rcmdcheck` to ensure Windows has both builds tested.

Because of #19 issues, Windows uses `stable-msvc` toolchain and pre-installed `MSYS2` (added to the `PATH`).

This PR is possible because I took the liberty of fixing my own error in PR #24.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -3,10 +3,12 @@ on:
     branches:
       - main
       - master
+      - updated-tests
   pull_request:
     branches:
       - main
       - master
+  workflow_dispatch:
 
 name: R-CMD-check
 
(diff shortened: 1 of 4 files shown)


yutannihilation (approved):
Looks good to me.

clauswilke:
Agreed, except `skip_on_cran()` is not the right approach at this time in my opinion. We'll have to manually add it to every set of tests we write, and if we ever want to test on CRAN we'll have to go in and remove all these calls. I think at this time one global check is more appropriate, and it should be added here:

https://github.com/extendr/rextendr/blob/fd64ee1f2fd9ab28bbbbfeb88703771ae76305a5/tests/testthat.R#L5-L8

Ilia-Kosenkov:
@clauswilke,
What kind of check do you have in mind?

clauswilke:
Just check the environment variable directly at the top level and don't execute the tests at all if it is not set. If you look at the testthat sources, their internal `on_cran()` function is a simple one-liner:
https://github.com/r-lib/testthat/blob/fe2bed2f3887b6d8ea84888f40d206edceb0d3ce/R/skip.R#L131

We should skip all tests if:
- We're on CRAN
- Rust is not available
- Maybe some other conditions, such as libR-sys doesn't exist for the current architecture. Let's leave that for the future.

clauswilke:
So something like:

```r
cargo_available <- system2("cargo") == 0L
not_cran <- identical(Sys.getenv("NOT_CRAN"), "true")
 
 if (cargo_available && not_cran) { 
   test_check("rextendr") 
 } 
```

yutannihilation:
I think rextendr needs something like `cargo_available()`, so that we can write

``` r
skip_if_not(cargo_available())
```

and also it's probably useful other places to provide user friendly errors or warnings.

c.f. https://github.com/rstudio/rmarkdown/blob/e848fc7a2d2fb500b4e1753172c5302eb58e446b/tests/testthat/test-lua-filters.R#L11

Ilia-Kosenkov:
@yutannihilation,
I totally agree. In this PR I just want to set up basic tests so any other changes to the codebase will get tested before merge, on multiple OS/architectures.

clauswilke:
@yutannihilation My reasoning is that at this point everything in the package requires a working Rust setup, and I have difficulty imagining what functionality we could add that wouldn't require one, so I prefer one global check. If this changes in the future I'm happy to reconsider.

clauswilke (approved):

PR #26, opened 2021-01-14
Default windows toolchain set to `gnu`

Ilia-Kosenkov:
In continuation to #25 and discussion in #19.

The current agreement is that if no `toolchain` is passed to R functions on Windows, we use `stable-x86_64-gnu` for 64 bit and `stable-i686-gnu` for 32 bit.
If however `toolchain` is provided, its value is used as-is, which allows overrides, e.g. `stable-msvc` with `MSYS2` linker.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -38,6 +38,12 @@ jobs:
         with:
           toolchain: ${{ matrix.config.rust-version }}
           default: true
+      
+      - name: Configure second Rust toolchain (Windows)
+        if: startsWith(runner.os, 'Windows') && endsWith(matrix.config.rust-version, 'gnu')
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable-i686-gnu
 
       # Uses @master branch to address rtools path issue
       # https://github.com/r-lib/actions/issues/228
@@ -65,10 +71,6 @@ jobs:
           key: ${{ runner.os }}-${{ hashFiles('.github/R-version') }}-1-${{ hashFiles('.github/depends.Rds') }}
           restore-keys: ${{ runner.os }}-${{ hashFiles('.github/R-version') }}-1-
 
-      - name: Configure Windows
-        if: startsWith(runner.os, 'Windows')
-        run: |
-            rustup toolchain add stable-i686-pc-windows-gnu
 
       - name: Configure Linux
         if: startsWith(runner.os, 'Linux')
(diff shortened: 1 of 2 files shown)


clauswilke:
@Ilia-Kosenkov I'm not sure I completely follow yet, and I think we should think this through a little more before proceeding:

> The current agreement is that if no toolchain is passed to R functions on Windows, we use stable-x86_64-gnu for 64 bit and stable-i686-gnu for 32 bit.

I think we should do everything we can to make sure the same Rust setup works for R packages with extendr and for rextendr calls. Currently we recommend msvc for R packages, and here you're recommending gnu for rextendr calls?

I don't have a preference either way, since I don't use Windows, but I do think we should recommend a single default toolchain.

Ilia-Kosenkov:
@clauswilke,
Let me summarise briefly what we found in discussion in #19.
When `*-msvc` is used, there is a need for external linker. This can come from `MSYS2` if `MSYS2` is on the `PATH` (both 32 and 64 bit). This was my setup by default and I saw no issues building and using `rextendr`.
However, `rextendr` can work without dependency on `MSYS2`.
For that, we need to use `*-gnu` toolchain. 
Unfortunately, toolchains should be different for 32- and 64- bits.

So, `rextenr` can work either using `*-msvc` and `MSYS2` or two `*-gnu` toolchains.
I am perfectly OK with the fact that we can make it work without extra dependency.

Perhaps we can go back and rewise how other `*extendr` packages are build with pre-defined bindings. Maybe `gnu` option will be a better choice.

clauswilke:
Thanks for the clarification. As I said before, I don't have a strong opinion on which way to go, but we should pick a single default tool chain for all cases.

Ilia-Kosenkov:
@clauswilke,
Ok I see.
Unfortuantely because of issues like https://github.com/rust-lang/cargo/issues/8990 there cannot be a simple and strighforward solution.

clauswilke:
Since things generally work with MSYS2, I would make that the default at this time. We just need to tell people that they need to install it and put it on their path. That's easy enough.

If at some point the setup without MSYS2 gets easier, we can switch to that being the default. But I wouldn't go there until we can reliably (and easily) build R packages with extendr dependency without MSYS2.

clauswilke:
One way to make sure we're doing something reasonable going forward is to address this issue: https://github.com/extendr/extendr/issues/127

Ilia-Kosenkov:
@clauswilke,
I agree that `MSYS2` solution works so far in all scenarios and I am not sure that `gnu` toolchain will be fixed.
Right now `rextendr` uses `stable-msvc` and `msys2` to run tests for both Windows architectures and it works fine.
I am closing this PR as at this point it is no longe relevant.

PR #27, opened 2021-01-16
Update setting_up_rust vignette about MSYS2 path

yutannihilation:
* Change MSYS2 example path to ~~`C\:msys2`~~ `C:\msys64`
* Change the PATH order; `mingw64` first and `mingw32` second
* Rebuild pkgdown site

### DIFF
--- a/docs/404.html
+++ b/docs/404.html
@@ -140,7 +140,7 @@ Content not found. Please use links in the navbar.
 </div>
 
 <div class="pkgdown">
-  <p>Site built with <a href="https://pkgdown.r-lib.org/">pkgdown</a> 1.6.0.</p>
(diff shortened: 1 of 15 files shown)


clauswilke:
Can you make it `msys64`, as described here: https://www.msys2.org/wiki/MSYS2-installation/?

yutannihilation:
Sorry, I didn't notice `C:\msys64` is the current default (and probably `C:\msys32` few years ago?)... In my environment msys2 is installed at `C:\msys2`, but I don't remember how I decided the path. As [GHA windows runner](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md#msys2) also uses `C:\msys64`, let's assume it's most typical one.

clauswilke (approved):

PR #29, opened 2021-01-30
Port over to new wrapper framework

clauswilke:
This PR switches over to the new wrapper framework implemented by @andy-thomason. It makes the code a little more complex but more robust and flexible.

Still to do:
- [x] update vignettes
- [x] rename `extendrfuns` chunk type, since it now will need a `extendr_module!` call? Maybe to `extendrsrc`?
- [x] rebuild website


Ilia-Kosenkov:
Ok, it looks good, but I have several moderate issues here.
First is the regex for extracting function names. I would suggest something like `\s*fn\s+(_\w+|[A-z]\w*)`, in line with what Rust defines as valid identifiers.
Second, do we really need to write down wrappers to a physical file? Perhaps we can make it opt-in for debug purposes and consume it within rextendr from in-memory?
Third, the need for `extendr_module!` is discouraging. Perhaps we can attempt to extract function names tracking `#[extendr]` macros?  I do not insist, but before this update, it was so nice to just write the necessary code with no boilerplate and enjoy the results. Even if this will be improved in the future on extendr site, we can temporarily provide the bypass in rextendr.

clauswilke:
> Third, the need for `extendr_module!` is discouraging.

Yes, that's why I was somewhat against this approach also. We can go back to using regular expressions to parsing the Rust code and finding the functions. I had written code for that in the past and didn't like it because it's not robust. But it may be the route we have to go, and we can improve our Rust parsing over time, and/or tell people to write somewhat reasonably formatted Rust.

clauswilke:
> Second, do we really need to write down wrappers to a physical file? Perhaps we can make it opt-in for debug purposes and consume it within rextendr from in-memory?

This is probably possible but low priority for me right now. I'll leave the source as is right now, but feel free to address after I merge.

Alternative, if there's a two-line solution you can point me to that I can drop into my code I'm happy to do so.

Ilia-Kosenkov:
I see your point. If we are switching to the new syntax that requires `extendr_module!`, could you please also write it down in the README so people don't get confused (and we don't forget).

clauswilke:
I've decided to make it optional. Default will be to parse the code with regex and auto-generate a module statement, but people will be able to turn it off and manually add their own, for more complex cases where the auto-generation fails. I think that's the best of both worlds.

Ilia-Kosenkov (approved):
The overall update looks good to me so merge as soon as you think it is ready.

clauswilke:
@Ilia-Kosenkov Could you review one more time? I've added code to generate the module macro.

For some reason the markdown integration is slightly broken now (output isn't captured, results aren't returned to R). I don't see why this code change would have caused that, though, so I'm a bit mystified.

clauswilke:
The markdown integration is a separate issue: #32

R/make_module_macro.R
```diff
@@ -0,0 +1,53 @@
...
+  outlines <- c(outlines, glue::glue_data(idents, "{type} {ident};"))
+  outlines <- c(outlines, "}")
+  outlines
+}
+
+
+get_extendr_idents <- function(lines) {
+  # find all lines with the "#[extendr]" decoration
+  idx <- which(grepl("\\#\\[extendr\\]", lines))
+  idx <- idx + 1 # bump indices by one to look in subsequent lines
```
Ilia-Kosenkov, line 29: Is it possible to have another attribute between `#[rextendr]` and function declaration?
I see that empty lines are dropped, but what about commented-out lines?
It does not make sense for Rust one-liners, but if you store some rust functions in a text file and send this text as an input, it may very well contain poorly placed comments.
```Rust
#[rextendr]
// Not good but still valid, right?
fn unmatched_fun() {}
```
If you feel like this may be important, perhaps we should identify commented-out lines as-well.

clauswilke: Ok, added code to strip all lines that are fully commented out. I could also add code to strip all lines that contain an attribute that is not `#[extendr]`, if you can think of a clever regex that identifies those lines. Otherwise I'll leave it as a TODO for the future.

Ilia-Kosenkov: @clauswilke,
I suggest leave it as todo to speed up things, as extra attributes are much rarer than just comments. 
We should also probably error if we are unable to find exports (i.e. malformed rust source), but this can be verified later.

I will play around with some regex and report to your todo issue.

Ilia-Kosenkov (approved):

PR #31, opened 2021-02-01
Configuring tests using environment variables

### LINKED ISSUE #30: Using environment variables to cofigure tests
Is it a good practice to use environment variables to configure tests? rextendr can be used for testing extendr (and in other situations), but for this, we need to configure `patch.crates_io` and, possibly, `toolchain`. 
This can be achieved using env vars, but I am not sure if this is a good practice.

Ilia-Kosenkov:
Continues discussion in and fixes #30.
The idea is to be able to control how {rextendr} builds rust code in tests using environment variables.
E.g., if tests should be executed using a different toolchain, use `REXTENDR_TOOLCHAIN`:
```R
Sys.setenv(REXTENDR_TOOLCHAIN = "stable-x86_64-pc-windows-gnu")
rcmdcheck::rcmdcheck(".")
```

Currently, only two variables are supported, `REXTENDR_TOOLCHAIN` and `REXTENDR_PATCH_CRATES_IO`.
Different crates patches can be separated using `';'`.

If the environment variable is not set (or it is an equivalent of an empty string), default values from `rust_source` are used.

### DIFF
--- a/tests/testthat.R
+++ b/tests/testthat.R
@@ -6,19 +6,27 @@ cargo_available <- system2("cargo") == 0L
 not_cran <- identical(Sys.getenv("NOT_CRAN"), "true")
 
 if (cargo_available && not_cran) {
+  # This retrieves exrepssions that represent
+  # default arguments of rust_source,
+  # which are used for reference
   test_env <- formals(rextendr::rust_source)
 
+  # Toolchain is a string, so can be read as is
   toolchain <- Sys.getenv("REXTENDR_TOOLCHAIN")
   if (!is.null(toolchain) && nzchar(toolchain)) {
     test_env$toolchain <- toolchain
     message(paste0(">> {rextendr}: Using toolchain from 'REXTENDR_TOOLCHAIN': ", toolchain))
   }
 
+  # Patch is represented as vector of character.
+  # In environment variable different crates are separated using ';'
+  # E.g., "extendr-api = { git = \"https://github.com/extendr/extendr\" };extendr-macros = { git = \"https://github.com/extendr/extendr\" }"
   patch <- Sys.getenv("REXTENDR_PATCH_CRATES_IO")
   if (!is.null(patch) && nzchar(patch)) {
     test_env$patch.crates_io <- strsplit(patch, ";")[[1]]
     message(paste0(">> {rextendr}: Using cargo patch from 'REXTENDR_PATCH_CRATES_IO': ", patch))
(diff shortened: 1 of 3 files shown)


tests/testthat.R
```diff
@@ -6,5 +6,29 @@ cargo_available <- system2("cargo") == 0L
...
+  # Patch is represented as vector of character.
+  # In environment variable different crates are separated using ';'
+  # E.g., "extendr-api = { git = \"https://github.com/extendr/extendr\" };extendr-macros = { git = \"https://github.com/extendr/extendr\" }"
+  patch <- Sys.getenv("REXTENDR_PATCH_CRATES_IO")
+  if (!is.null(patch) && nzchar(patch)) {
+    test_env$patch.crates_io <- strsplit(patch, ";")[[1]]
+    message(paste0(">> {rextendr}: Using cargo patch from 'REXTENDR_PATCH_CRATES_IO': ", patch))
+  } else {
+    # Patch should be evaluated because it can be represented as a vector
+    test_env$patch.crates_io <- eval(test_env$patch.crates_io)
```
clauswilke, line 30: Not sure I understand this line. Shouldn't `test_env$path.crates_io` be `NULL` in this branch? If not, maybe some additional comments are needed to explain?

clauswilke: Ah, I think I get it now. Never mind.

tests/testthat.R
```diff
@@ -6,5 +6,29 @@ cargo_available <- system2("cargo") == 0L
 not_cran <- identical(Sys.getenv("NOT_CRAN"), "true")
 
 if (cargo_available && not_cran) {
+  # This retrieves exrepssions that represent
+  # default arguments of rust_source,
+  # which are used for reference
+  test_env <- formals(rextendr::rust_source)
```
clauswilke, line 12: Not sure I like the name `test_env`. It sounds more general than it is. How about `rust_source_defaults`?

Ilia-Kosenkov:
@clauswilke,
So what is your opinion on this? Should I proceed?

clauswilke:
Sure, go ahead.

clauswilke:
For some reason, my local checks fail now with `object 'rust_source_defaults' not found`. Not sure what is happening.

But also, now that I think about it, as it is currently implemented this is somewhat problematic, as it now requires every additional test to set the `toolchain` and `patch.crates_io` arguments. We may have to rethink this.

Issue #35, opened 2021-02-04
`rust_source` overhaul


Ilia-Kosenkov:
In light of recent PRs (#29, #31, #34) I would like to suggest an overhaul of `rust_source` function that will simplify its usage, especially in the test environment.

- [x] #44 Problematic parameters
 `rust_source` has a number of parameters that are very difficult to provide manually. I suggest creating a function (or use some package) that will convert `list` into toml-compatible representation. This will allow us to replace (at least) the following parameters
     a. `dependencies = list(rand = "0.8.3", syn = "1.0.60")`
     b. ``` patch.crates_io = list(`extendr-api` = list( git = "..."), `extendr-macros` = list( git = "...")) ```
I understand objections to `list` syntax, but it will allow usage of `path`  or `branch` parameters, helpful while testing `{extendr}`/`{rextendr}`

- [x]  #36 `options` for default values.  
Instead of supplying complex default values, we should use `options`. I would suggest at least the following default options:
     a. `options(rextendr.dependencies = NULL)`
     b. `options(rextendr.toolchain = NULL)`
     c. `options(rextendr.patch.crates_io = list(...))`
This will allow us updating function paramters to `rust_source(..., patch.crates_io = getOption("rextendr.patch.crates_io"))`. 
Default options can be also chained with other patches simply using `append(getOption("rextendr.patch.crates_io"), list(my_crate = list(path = "...")))`

- [x] #38 Referencing RTools on Windows
When we build regular extendr-based packages, RTools gets included in the search path by R build tools.
When we dynamically build rust libs using `{rextendr}`, no magic happens and the regular path gets used. By convention, RTools does not recommend exposing its `mingwXX/bin` paths to system's `PATH`, so it may not be present there. This creates testing problems due to the cross-compilation on Windows -- one cannot run `rcmdcheck` without first setting up system `PATH`. The inclusion of RTools can also be configured using `options()`.
I suggest to move `system2` cargo call to a separate function and add the following preamble to it:
<details>
<summary> invoke_cargo() </summary>

```R
 # Append rtools path to the end of PATH
  if (.Platform$OS.type == "windows" && nzchar(Sys.getenv("RTOOLS40_HOME"))) {
    env_path <- Sys.getenv("PATH")
    r_tools_path <-
      normalizePath(
        file.path(
          Sys.getenv("RTOOLS40_HOME"), # {rextendr} targets R >= 4.0
          paste0("mingw", ifelse(R.version$arch == "i386", "32", "64")),
[...]

PR #36, opened 2021-02-08
Using `options` to configure `rust_source`

Ilia-Kosenkov:
Using `options` to provide default values for `rust_source`, as discussed in #35.


R/zzz.R
```diff
@@ -1,9 +1,38 @@
...
+    rextendr.patch.crates_io =  c(
+      'extendr-api = { git = "https://github.com/extendr/extendr" }',
+      'extendr-macros = { git = "https://github.com/extendr/extendr" }'
+    ),
+    # character scalar
+    # Default version of 'extendr_api' if no 'patch.crates_io' is specified
+    rextendr.extendr.version = "*",
+    # character scalar
+    # Default version of 'extendr_macros' if no 'patch.crates_io' is specified
+    rextendr.extendr_macros.version = "*"
```
clauswilke, line 31: I'm a bit unclear about when to use `.` vs `_` as a separator but I guess it's fine the way you set it up.

clauswilke (approved):

R/zzz.R
Ilia-Kosenkov: @clauswilke,
From what I observed in options present on my system and set by other packages -- `'.'` is used as a separator, e.g. `{ggplot2}` https://github.com/tidyverse/ggplot2/blob/9741da5050f81b7b5c012c56d02f45fc93d68f89/R/scale-colour.r#L43-L45.
So the naming convention looks like `<package>.<global_option>` and `<package>.<entity>.<property>`.
Usage of `'_'` is adopted from rextendr itself, where it replaces `'-'`, e.g. `patch.crates_io` is an equivalent of Cargo.toml's `[patch.crates-io]`. Same for `extendr-*`.

Ilia-Kosenkov:
@clauswilke,
I have added simple tests to check if `options` have any effect on `rust_source` (by supplying invalid option values and runnig `rust_function`, expecting Rust compilation error).
If there are no issues, I can merge this (after tests finish once again).

clauswilke:
Yes, looks good.

clauswilke:
Oh, you should also add your name to the authors list in `DESCRIPTION`, at the end.

Ilia-Kosenkov:
@clauswilke,
I have another PR ready, I will do it in the next one, thank you.
Merging this now.

PR #38, opened 2021-02-08
Appending Rtools path on Windows

Ilia-Kosenkov:
When extendr is built as part of a project, Rtools are injected by the build system. 
However, rextendr builds libraries at runtime and is unaware of Rtools. This PR appends Rtools path (architecture-dependent) to the system's PATH for the duration of cargo build if `RTOOLS40_HOME` environment variable is set. This is optional and is controlled by the logical parameter `use_rtools` (`TRUE` by default).


R/source.R
```diff
@@ -215,6 +207,49 @@ rust_function <- function(code, env = parent.frame(), ...) {
   rust_source(code = code, env = env, ...)
 }
 
+invoke_cargo <- function(
+  toolchain, specific_target,
+  dir, profile, stdout, stderr,
+  use_rtools) {
```
clauswilke, line 213: Function arguments are formatted a bit strangely here. We normally follow the format you see e.g. in `generate_cargo.toml()` on lines 253/254.

R/source.R
```diff
@@ -34,6 +34,10 @@
 #' @param cache_build Logical indicating whether builds should be cached between
 #'   calls to [rust_source()].
 #' @param quiet Logical indicating whether compile output should be generated or not.
+#' @param use_rtools Logical indicating whether to append path to Rtools
+#'   to the PATH variable on Windows using RTOOLS40_HOME env var (if it is set).
```
clauswilke, line 38: I think `PATH` and `ROOLS40_HOME` should be in backticks.

R/source.R
```diff
@@ -215,6 +207,49 @@ rust_function <- function(code, env = parent.frame(), ...) {
   rust_source(code = code, env = env, ...)
 }
 
+invoke_cargo <- function(
+  toolchain, specific_target,
+  dir, profile, stdout, stderr,
+  use_rtools) {
+  # Append rtools path to the end of PATH on Windows
+  if (
+    use_rtools &&
```
clauswilke, line 216: Since this is a parameter a user can set, you should write `isTRUE(use_rtools)`.

R/source.R
```diff
@@ -215,6 +207,49 @@ rust_function <- function(code, env = parent.frame(), ...) {
...
+      normalizePath(
+        file.path(
+          Sys.getenv("RTOOLS40_HOME"), # {rextendr} targets R >= 4.0
+          paste0("mingw", ifelse(R.version$arch == "i386", "32", "64")),
+          "bin"
+        )
+      )
+    Sys.setenv(PATH = paste(env_path, r_tools_path, sep = .Platform$path.sep))
+    # This retores PATH when function returns, i.e. after cargo finishes.
+    on.exit(Sys.setenv(PATH = env_path))
```
clauswilke, line 231: Better to set the exit handler before you modify the `PATH` variable, instead of after.

R/source.R
```diff
@@ -149,24 +155,10 @@ rust_source <- function(file, code = NULL,
...
-      if (!is.null(specific_target)) sprintf("--target %s", specific_target) else NULL,
-      sprintf("--manifest-path %s", file.path(dir, "Cargo.toml")),
-      sprintf("--target-dir %s", file.path(dir, "target")),
-      if (profile == "release") "--release" else NULL
-    ),
-    stdout = stdout,
-    stderr = stdout
+  invoke_cargo(
+    toolchain, specific_target, dir,
+    profile, stdout, stderr, use_rtools
```
clauswilke, line 160: When there are so many arguments, it's generally better to name them explicitly to avoid difficult-to-debug problems with argument order.

Ilia-Kosenkov:
@clauswilke,
I think I addressed all the problems you mentioned.

clauswilke (approved):

PR #39, opened 2021-02-08
Add pkgdown action

malcolmbarrett:
This PR sets up a GitHub Action to automatically rebuild and redeploy the pkgdown site whenever pushing to the primary branch. It also removes the `docs/` folder that contained a manually built version.

You'll likely need to tweak the GH Pages settings to use the`gh-pages` branch instead

Also, this does not reckon with trying to re-knit the readme, although I suppose it could.


clauswilke:
Could you leave the manually compiled site in until we know that we can deploy from the branch? If I switch over now the site is gone.

malcolmbarrett:
Good point. I'll update

clauswilke:
And you'll probably have to add a Rust setup for this to work:
https://github.com/extendr/rextendr/blob/b8afaf17ee88644489a19ffde961aa213c922b04/.github/workflows/R-CMD-check.yaml#L35-L39

malcolmbarrett:
Oh right! Both done.

malcolmbarrett:
wait, that toolchain part is probably not right

malcolmbarrett:
This is on mac, so should it be...?

```yaml
with:
    toolchain: stable
    default: true
```

clauswilke:
Yes, I think so.

malcolmbarrett:
Ok, updated

clauswilke:
Ok, let's merge and see what happens.

clauswilke:
Seems to have worked. I switched over  to gh-pages and the site appears to be fine.

yutannihilation:
:tada: 

@clauswilke 
It seems there's no link to the pkgdown site. Could you add the link to <https://extendr.github.io/rextendr/> on the repository's description? (or on the README?)

clauswilke:
@yutannihilation Could you file an issue? I'm super busy this week and may forget.

yutannihilation:
Sure.

malcolmbarrett:
Great! Checked the action and everything looks good.

PR #42, opened 2021-02-09
Cleanup DESCRIPTION file

malcolmbarrett:
This PR runs two usethis functions: `use_github_links()` and `use_tidy_description()` to clean up the DESCRIPTION file a bit. My primary motivation was to activate links to GitHub in the pkgdown site but I figured it could use a good tidy while I was at it


clauswilke:
Thanks!

PR #44, opened 2021-02-09
Toml serialization

Ilia-Kosenkov:
As was discussed in #35, currently one of the arguments of the `rust_source`, namely `patch.crates_io` accepts rather complicated values as strings. It may be beneficial to use structured notation based on e.g. R `list`s instead of plain strings. Because lists allow nesting and have (optional) item names, they are the best choice. The goal *was* to be able to take
```R
list(
    `extendr-api` = list( git = "https://github.com/extendr/extendr"),
    `extendr-macros` = list(git = "https://github.com/extendr/extendr")
)
```
and produce
```toml
extendr-api = { git = "https://github.com/extendr/extendr" }
extendr-macros = { git = "https://github.com/extendr/extendr" }
```

Using the sample I posted in #35 I made a tiny serializer using `S3` dispatch.
It turned out that this serializer can now be used to produce full Cargo.toml files (in the majority of cases, as Cargo.toml uses a subset of toml standard).
Basically, now we *can have* (mimicking https://github.com/extendr/extendr/blob/master/extendr-engine/Cargo.toml):
<details>
<summary>Invoking from R</summary>

```R
to_toml(
  package = list(
    name = "extendr-engine",
    version = "0.1.11",
    authors = c("author_1", "author_2", "author_3"),
    edition = "2018",
    description = "Safe and user friendly bindings to the R programming language.",
    license = "MIT",
    repository = "https://github.com/extendr/extendr"
  ),
  dependencies = list( `libR-sys` = "0.2.0"),
  features = list(
    default = character(0),
    `tests-all` = array("libR-sys/use-bindgen", 1)
  )
)
```
</details>

<details>
<summary>which produces this</summary>

```toml
[package]
name = 'extendr-engine'
version = '0.1.11'
authors = [ 'author_1', 'author_2', 'author_3' ]
edition = '2018'
description = 'Safe and user friendly bindings to the R programming language.'
license = 'MIT'
repository = 'https://github.com/extendr/extendr'
[dependencies]
libR-sys = '0.2.0'
[features]
default = [ ]
tests-all = [ 'libR-sys/use-bindgen' ]
```

</details>

Pros: 
  - Rudimentary validation and error handling
  - Type-awareness, correct string and numeric formatting
  - Arbitrary structure as long as it uses valid syntax
  - R validates list creation syntax (extra safety compared to no safety in case of strings)
  - Unifies all Cargo.toml - related arguments
  - Works nicely with `options()` to provide default values

Cons:
  - Verbose
  - May be unable to produce arbitrary toml, but should be enough for auto-generation
  - Requires extra dependencies: mainly `rlang`, which was previously imported implicitly
  - Complicates function syntax
  - Complicates interaction with the environment variables (can be solved using at least two approaches)
  - Should not break CI & test pipelines, but cannot be 100% sure
  - Changes API surface through `rust_source`
  - ???


clauswilke:
I like this in principle. A couple of comments:

- Please use tidyverse styling. This applies in particular to indentation (2 spaces) and formatting of function arguments. See in particular 2.4.1 and 3.2 here: https://style.tidyverse.org/
- Do we actually need to state the `extendr-macros` dependency? I believe we don't. See here: https://github.com/extendr/extendr/blob/ee4fa13e6988a39af13abaded1d2b6d80c13d1d2/tests/extendrtests/src/rust/Cargo.toml#L15-L22
- Any reason you enclosed the examples into `\dontrun{}`? They don't require Rust and they're not slow, so there shouldn't be a reason not to run them. Then you can also remove the comments providing the output, as it will be generated automatically for the website.

Ilia-Kosenkov:
@clauswilke,
Thank you, I am glad this PR looks reasonable to you. 

1. Sure I will do that.
2. I do not know, I did not touch any rextendr core code and worked only with the public API. I will change the default value in options and see how tests go.
3. My mistake, I used docs from another method as a template.

Ilia-Kosenkov:
@clauswilke,
I used `{styler}` to fix existing code (files that I have touched) and I threw away `extendr-macros` dependency. Docs were also updated.

R/source.R
```diff
@@ -255,39 +262,37 @@ invoke_cargo <- function(toolchain, specific_target, dir, profile,
...
-  # add user-provided dependencies
-  cargo.toml <- c(cargo.toml, dependencies)
-
-  # add user-provided patch.crates-io statements
-  cargo.toml <- c(
-    cargo.toml,
-    "[patch.crates-io]",
-    patch.crates_io
+generate_cargo.toml <- function(
+                                libname = "rextendr",
```
clauswilke, line 266: The first argument should be on the same line as `function(`. Does styler write it like this?

R/toml_serialization.R
```diff
@@ -0,0 +1,219 @@
+#' Convert R `list()` into toml-compatible format.
+#'
+#' [to_toml()] can be used to build `Cargo.toml`.
```
clauswilke, line 3: Could you expand the description to 2-3 sentences? This seems very terse.

R/toml_serialization.R
```diff
@@ -0,0 +1,219 @@
...
+#' @param .str_as_literal Logical indicating wether to treat
+#'     strings as literal (single quotes no escapes) or
+#'     basic (ecaping some sequences) ones. Default is `TRUE`.
+#' @param .format_int,.format_dbl Character scalar describing
+#'     number formatting. Compatible with `sprintf`.
+#' @return A character vector, each element corresponds to
+#'     one line of the resulting output.
+#' @examples
+#' to_toml(workspace = )
+#' # Produces [workspace] with no children
```
clauswilke, line 16: Please put the comments above the code, not below.

R/toml_serialization.R
```diff
@@ -0,0 +1,219 @@
...
+#'
+#' to_toml(patch.crates_io = list(`extendr-api` = list(git = "git-ref")))
+#' # [patch.crates_io]
+#' # extendr-api = { git = 'git-ref' }
```
clauswilke, line 20: Remove the output. It'll be generated automatically when the website is built. See e.g. here: https://dplyr.tidyverse.org/reference/mutate.html#examples

R/toml_serialization.R
```diff
@@ -0,0 +1,219 @@
...
+#'
+#' # Single-element arrays are distinguished from scalars
+#' # using explicitly set `dim`
+#' to_toml(lib = list(`crate-type` = array("cdylib", 1)))
+#' # [lib]
+#' # crate-type = [ 'cdylib' ]
```
clauswilke, line 26: As above.

R/toml_serialization.R
```diff
@@ -0,0 +1,219 @@
...
+#' @export
+to_toml <- function(
+                    ...,
```
clauswilke, line 29: As above, no newline after `function(`.

clauswilke (approved):
Overall looks good, but I've got a few more minor comments. In addition, consider using the various `glue` functions instead of `paste0()`. The result will generally be more readable.

CGMossa:
I hope this is not completely out of place: I recently found that one can provide a `[patch.crates-io]` so
that crates doesn't fetch the crates-version, but the "custom" version that is specified. See the [Replacing dependencies with patch in the Edition Guide](https://doc.rust-lang.org/stable/edition-guide/rust-2018/cargo-and-crates-io/replacing-dependencies-with-patch.html).

malcolmbarrett:
For the test, you might want to think about using `expect_snapshot_output(toml)` rather than manually typing/escaping the TOML. It's nice for these kinds of tests. That would require testthat 3.0.0, I believe, and may also require you to set the 3rd edition (`usethis::use_testthat(3)`)

Ilia-Kosenkov:
@malcolmbarrett,
Thanks, I was unaware of this feature.
The question is whether we want to depend on the 'experimental' functions.
I think it is up to @clauswilke to decide.

My only problem is that `expect_snapshot*` test against pre-existing snapshots, generated previously by the tested code. This is useful to verify if the new build produces the **same** output, but not that the output is **correct**.
Perhaps we should still have a reference .toml file (stored as a text file, not in R source code) that we test against? Is there a `{testhat}` solution for this?

clauswilke:
I like the way the test is currently set up, because it makes it very easy to manually compare the test input to the expected output. If you save the created toml file somewhere then you have different parts of the test in different files.

Ilia-Kosenkov:
@clauswilke,
Ok, if there are no other issues, I think we should merge.
I switched to string interpolation where applicable, fixed comments & formatting, and also added support for logicals and limited support for so-called array-of-tables, which effectively mean we should be able to produce any reasonable cargo manifest directly from R, no hand-picking or string editing required.
Toml test now reflects this change through the usage of `data.frame`.

clauswilke:
Fine by me.

Ilia-Kosenkov:
@clauswilke,
Just noticed & fixed a few typos, merging after the tests pass.

PR #47, opened 2021-02-16
Better parsing of rust code

### LINKED ISSUE #33: Make module generation more robust
The module generation code expects `#[extendr]` immediately in front of a function or impl declaration. It would be good to make it more robust, e.g., allow additional attributes between `#[extendr]` and the declaration.

Also possibly generate a warning of no `#[extendr]` attributes are found.

Ilia-Kosenkov:
Fixes #33.
The previous 'parser' of rust header did not take into account edge cases, such as comments, unusual line formatting, etc.
This is an attempt to improve the situation.
The goal is to be able to parse
1. Multiple attributes, i.e.
```Rust
#[extendr]
#[allow(dead_code)]
fn fn_1() {}
```
2. One-liners
```#[extendr] fn fn_2() {}```
3. Implementations with lifetimes
```#[extendr] impl<'a> X<'a>```
4. Strip line comments starting with `//`
5. Handle block (`/* .. */`) comments, like
```Rust
#  [/*comment*/extendr/*comment*/]
/* multi
line
comment*/ fn
fn_3(){}
```

While the majority of these cases are extremely rare and unlikely to appear in {rextendr}, it is good to be able to handle it.
Now we can also output better error messages if `#[extendr]` attribute is not attached to something exportable, or if comments are placed incorrectly.


R/clean_code.R
```diff
@@ -0,0 +1,151 @@
+utils::globalVariables(
+  c("cnt", "start", "impl", "lifetime", "type", "name")
+)
```
Ilia-Kosenkov, line 3: I use {dplyr} & {tibble} to ease data manpulation, so I obviously get R CMD check complaints about undeclared variables.
I invoked `globalVariables` here, but if you prefer another solution to this problem, I will gladly implement it.

Ilia-Kosenkov:
Ok so `4.0.4` was released today, and for some reason, `ubuntu-release` fails to download R `4.0.4`. Perhaps something with the CDN or there is a bug in r-actions.

R/clean_code.R
clauswilke: See if you can resolve the issue using pronouns: https://rlang.r-lib.org/reference/tidyeval-data.html

R/clean_code.R
```diff
@@ -0,0 +1,151 @@
...
+
+clean_rust_code <- function(lines) {
+  lines <-
```
clauswilke, line 6: Not sure why you make an assignment here. It's not needed as far as I can see.

R/clean_code.R
```diff
@@ -0,0 +1,151 @@
...
+# Because R does not allow strightforward iteration over
+# scalar strings, determining `/*` and `*/` positions can be challenging.
+# E.g., regex matches 3 `/*` and 3 `*/` in `/*/**/*/`.
+# 1. We find all occurence of `/*` and `*/`.
+# 2. We find non-overlapping `/*` and `*/`.
+# 3. We build pairs of open-close comment delimiters by collapsing nested
+#   comments.
+# 4. We fill in space between remaining delimiters with spaces (simplest way).
+fill_block_comments <- function(lns, fill_with = " ") {
+  lns <- paste(lns, collapse = "\n")
```
clauswilke, line 31: Let's use `glue` functions consistently instead of `paste()`, `paste0()`, etc. This here would probably be `glue_collapse()`.

tests/testthat/test-make-module-macro.R
```diff
@@ -55,3 +58,33 @@ impl Counter {
...
+  expect_error(
+    make_module_macro("*/  /*"),
+    paste(
+      "Malformed comments.",
+       "x `/\\*` and `\\*/` are not paired correctly.",
+       "i This error may be caused by a code fragment like `\\*/ ... /\\*`.",
+      sep = "\n "
+    )
+  )
+})
```
clauswilke, line 90: Please add a newline at the end.

clauswilke (commented):
I flagged a few minor issues. I didn't read through the logic in detail.

One bigger item, though: Since you're adding multiple helper functions you should add extensive unit tests for those also.

Ilia-Kosenkov:
@clauswilke,
Thanks, rlang pronouns indeed helped, I completely forgot about that feature.
I addressed other issues as well: removed `paste*` in favor of `glue::*` and fixed other minor problems.

clauswilke (approved):
Yes, looks good. Just add some unit tests for the various helper functions you added.

Ilia-Kosenkov:
@clauswilke,
I have added some tests, but I am not sure how useful they are. 
We test a lot of situations by testing public api `make_module_macro`, including thrown errors and in-text comments.
Still, I hope I covered the majority of problematic edge cases.
If there are no objections, I am ready to proceed.

clauswilke:
It's a general practice of good software engineering to write unit tests for every helper function. It's a way of documenting what the function does and making sure it continues working. And if something breaks, you'll quickly know where in the code the problem is.

Also, you probably wrote little tests for these functions, so just stick them into the test module. See also: https://twitter.com/hadleywickham/status/1360584935829086213?s=20

You can merge whenever you think it's ready.

PR #49, opened 2021-02-19
Fixing bug in toml serialization

Ilia-Kosenkov:
I found a small bug in toml serialization of [array-of-tables](https://toml.io/en/v1.0.0#array-of-tables), when R's `data.frame` has only one column, e.g. `data.frame(x = 1)`.
This is caused by inconsistent (in my opinion) behavior of `[i, ]`, which in the case of one column returns not the data.frame row (as named list), but rather the contents of the column (which I honestly never noticed).
```R
data.frame(x = 1, y = 1)[1, ]
x y
1 1
data.frame(x = 1)[1, ]
[1] 1
```

My choice of `[,]` operator was bad, so I suggest to replace it with `dplyr::slice`, which does the job by internally calling `vctrs::vec_slice`, which is type- and size-stable.

I have added extra tests to verify the code works in the edge cases.

### DIFF
--- a/R/toml_serialization.R
+++ b/R/toml_serialization.R
@@ -131,7 +131,6 @@ format_toml.data.frame <- function(x,
                                    .top_level = FALSE) {
   rows <- nrow(x)
   header <- glue("[[{.tbl_name}]]")
-
   if (rows == 0L) {
     return(as.character(header))
   }
@@ -139,7 +138,7 @@ format_toml.data.frame <- function(x,
     map(
       seq_len(rows),
       function(idx) {
-        item <- simplify_row(x[idx, ])
+        item <- simplify_row(dplyr::slice(x, idx))
         if (length(item) == 0L) {
           result <- character(0)
         } else {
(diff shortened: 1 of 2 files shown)


malcolmbarrett:
FYI you can also just do `data.frame(x = 1)[1, , drop = FALSE]`

clauswilke (approved):
Yes, that's a very common problem. The base R solution is `x[idx, , drop = FALSE]`, but it's a good idea to avoid indexing altogether for these reasons.

Ilia-Kosenkov:
@clauswilke, @malcolmbarrett,
Thank you for the clarification.
I originally used `[, ]` in attempt to fully support toml array of tables, for which we had to iterate over each value of data.frame, and I tried doing this and then forgot to update this code fragment.

I'd like to avoid indexing, but right now I see no simple way of row-wise indexing. I believe, this problem will be solved as soon as `vec_map` and friends are available. I can think of `dplyr::mutate/transmute` or `purrr::pmap`, but there is not much benefit, `to_toml` is called infrequently and usually processes very small objects.

So for now I'd like to merge this as-is and keep an eye on {vctrs} releases.

clauswilke:
Sure, go ahead.

PR #50, opened 2021-02-21
Use extendr-api from crates.io

clauswilke:
Now that extendr 0.2 is on crates.io, we should not use the github version any longer. Going forward, the goal should be for rextendr to always work with the extendr version on crates.io.


Ilia-Kosenkov (approved):
Looks good, please go ahead.

PR #52, opened 2021-03-04
add wrapper generation for packages

clauswilke:
Add a function `register_extendr()` that retrieves the wrapper code for a package that uses extendr. The wrappers are written to `R/extendr-wrappers.R`.


R/register_extendr.R
```diff
@@ -0,0 +1,57 @@
...
+      glue::glue("Package {pkg_name} cannot be loaded. No wrapper functions were generated."),
+      call. = FALSE
+    )
+  }
+}
+
+make_wrappers <- function(module_name, package_name, outfile,
+                          use_symbols = FALSE, quiet = FALSE) {
+  wrapper_function <- glue::glue("wrap__make_{module_name}_wrappers")
+  wrapper_call <- glue::glue(
```
yutannihilation, line 41: Can't we simply call .Call() here?

clauswilke: No, if we try that then R CMD check complains about a non-existent symbol.

R/register_extendr.R
```diff
@@ -0,0 +1,57 @@
...
+#' @export
+register_extendr <- function(path = ".", quiet = FALSE) {
+  x <- desc::desc(rprojroot::find_package_root_file("DESCRIPTION", path = path))
+  pkg_name <- x$get("Package")
+
+  if (!isTRUE(quiet)) {
+    message(glue::glue("Generating extendr wrapper functions for package: {pkg_name}"))
+  }
+
+  outfile = rprojroot::find_package_root_file("R", "extendr-wrappers.R", path = ".")
```
Ilia-Kosenkov, line 26: Should it be `<-` instead of `=`?

R/register_extendr.R
yutannihilation: Sorry, I couldn't reproduce the complain. There's no NSE involved here so I believe there's no non-existent symbols. Am I missing something...?

commit: https://github.com/yutannihilation/rextendr/pull/1/commits/bd5b63e0565a1431c32ccd7a281a8273be4ee65f
result: https://github.com/yutannihilation/rextendr/pull/1/checks

R/register_extendr.R
clauswilke: Yes, thanks for catching.

R/register_extendr.R
clauswilke: Strange. I am certain I tried this without `eval()` first, but your code seems to work. I don't remember what I wrote that didn't work.

yutannihilation: Maybe you used a symbol at that time? Anyway good that it works :)

yutannihilation (approved):
Looks good.

PR #53, opened 2021-03-04
Add function to create extendr package infrastructure

clauswilke:
I have added a function `use_extendr()` to add an extendr scaffolding to an R package. There is currently no unit testing for this, it's a bit difficult to set up.

@Ilia-Kosenkov Could you try this on Windows?


R/use_extendr.R
```diff
@@ -0,0 +1,170 @@
...
+"
+  )
+  brio::write_lines(makevars_win_content, file.path(src_dir, "Makevars.win"))
+
+  cargo_toml_content <- glue::glue(
+r"(
+  [package]
+  name = "{pkg_name}"
+  version = "0.1.0"
+  edition = "2018"
```
yutannihilation, line 110: How can users choose edition? I guess we'll want to choose 2021 here.

R/use_extendr.R
```diff
@@ -0,0 +1,170 @@
...
+{roxcmt} @usage NULL
+{roxcmt} @useDynLib {pkg_name}, .registration = TRUE
+NULL
+
+{roxcmt} Return string `"Hello world!"` to R.
+{roxcmt} @export
+hello_world <- function() .Call(wrap__hello_world)
+)"
+  )
+  brio::write_lines(wrappers_content, wrappers_file)
```
yutannihilation, line 163: Can we leave it to users to run `register_extendr()` instead of manually inserting these lines? I'm afraid we'll forget to fix here when we make some changes on wrapper generation...

yutannihilation (commented):
Do we add `src/.gitignore`?


```
*.o
*.so
*.dll
target
Cargo.lock
```
https://github.com/extendr/helloextendr/blob/main/src/.gitignore

R/use_extendr.R
clauswilke: Is 2021 out already?

R/use_extendr.R
clauswilke: Users won't be able to compile unless they have the `@useDynLib` line. Therefore, I thought it would be best to add a wrapper file right away. It'll get overwritten the moment they rerun `register_extendr()`, so it only has to be approximately correct. (In fact, it just has to be not so wrong that the package won't build.)

clauswilke:
I can add `src/.gitignore`. There's some discussion on whether `Cargo.lock` should or should not get version tracked, (don't track for Rust library crates, track for enduser applications) so maybe leave out of default `.gitignore`?

R/use_extendr.R
yutannihilation: Not yet. Will we just replace this to `2021` when it's out? If so, I'm fine with no options.

R/use_extendr.R
yutannihilation: I see. Fair enough.

yutannihilation:
Ah, I agree we should leave `Cargo.lock`, but I think `taget` should be ignored in any case.

yutannihilation:
Also, these are not included in `.gitignore` if I create a package on RStudio. What do you mean by "default `.gitignore`"?

```
*.o
*.so
*.dll
```

R/use_extendr.R
clauswilke: Yes, I think that's the right way to go.

clauswilke:
> What do you mean by "default `.gitignore`"?

I was looking here, since I'm copying helloextendr basically: 
https://github.com/extendr/helloextendr/blob/main/src/.gitignore

clauswilke:
`.gitignore` added.

R/use_extendr.R
yutannihilation: Okay.

yutannihilation:
Thanks, looks good now. Let me try creating a package using this before approving.

Ilia-Kosenkov:
@clauswilke,
If it is OK, please delay merging a little bit - I'd like to test a few things.

andy-thomason:
Very nice. I fully intend to use this to generate some example packages
as soon as I am able.

clauswilke:
@Ilia-Kosenkov Sure, no problem.

Ilia-Kosenkov (commented):
@clauswilke,
It works fine under Windows.
However, I suggest using `to_toml` to generate the skeleton `Cargo.toml` (see comments).

R/use_extendr.R
```diff
@@ -0,0 +1,177 @@
...
+  version = "0.1.0"
+  edition = "2018"
+
+  [lib]
+  crate-type = ["staticlib"]
+
+  [dependencies]
+  extendr-api = "*"
+)"
+  )
```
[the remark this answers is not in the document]
clauswilke: Thanks, will do.

Ilia-Kosenkov:
@clauswilke,
I was thinking about adding a test CI step that generates a dummy package, uses `use_extendr()` and then runs `rcmdcheck` on a simple test that calls `hello_world()`. If everything works as expected, the R CMD check will be successful.
Would you consider adding it here or should we discuss it in a separate issue/pr?

clauswilke:
@Ilia-Kosenkov I think it's easier if I merge this and then you make a separate PR for the test.

Ilia-Kosenkov (approved):
Sure, go ahead.

Issue #56, opened 2021-03-06
Can we generate extendr-wrappers.R at the same time of compiling Rust code?


yutannihilation:
As described on [helloextendr's README](https://github.com/extendr/helloextendr#generate-wrappers), the current steps to produce the wrapper are:

``` r
# Compile the Rust code (If you are using RStudio, just run "Install and Restart")
devtools::install()

# Re-generate wrappers
rextendr::register_extendr()

# Re-generate documentation and NAMESPACE (If you are using RStudio, just run "Document")
devtools::document()
```

But, I come to wonder if explicit `rextendr::register_extendr()` is really needed. Can we generate it on compilation by tweaking some `build.rs`? (Or `Makevars`?)

yutannihilation:
> by tweaking some `build.rs`?

Apparently, it cannot because `build.rs` is what is executed before compilation.

But, `Makevars` works. It's just creating `main.rs` like this and do `cargo run` in `Makevars`. Can we switch to this?

``` rust
fn main() {
    let metadata = helloextendr::get_helloextendr_metadata();
    print!(
        "{}",
        metadata.make_r_wrappers(true, "helloextendr").unwrap()
    );
}
```

full commit: https://github.com/yutannihilation/helloextendr/commit/e556d81e3dd846d9f51f3ed5636a1cb931632f8d

yutannihilation:
@clauswilke @andy-thomason @Ilia-Kosenkov 
Sorry I think I didn't follow the discussion on the wrapper-generation completely, but are there any reason the wrapper generation needs to be called from R session? The `Makevars`-method I described above seems to work, and I'm wondering if there's something I misunderstand.

clauswilke:
Yes, your `Makevars` method should work also. It's just a matter of preference. Do you want more boilerplate Rust code or do you want an extra R call?

Also, have you tested your method on Windows? Didn't we have problems with `cargo run` there or has that been solved? And on Windows, we'll generate the wrappers twice, but maybe that's not a big deal.

yutannihilation:
> It's just a matter of preference. Do you want more boilerplate Rust code or do you want an extra R call?

I don't think this is just my preference. This matters because if this works the current steps:

``` r
# Compile the Rust code (If you are using RStudio, just run "Install and Restart")
devtools::install()

# Re-generate wrappers
rextendr::register_extendr()

# Re-generate documentation and NAMESPACE (If you are using RStudio, just run "Document")
devtools::document()
```

[...]

PR #57, opened 2021-03-07
Implement force_wrappers argument of register_extendr()

yutannihilation:
As suggested here: https://github.com/extendr/rextendr/issues/56#issuecomment-792186856


clauswilke (approved):
It's generally fine. I'm just not sure about the word "sample" in `sample_function_wrapper()` etc. What is it supposed to mean? Do you mean "example"?

clauswilke:
The other thing I've been wondering is whether we should check for the presence of `src` and `src/entrypoint.c` and execute `register_wrappers()` only if they exist? As a way to check whether there actually is compiled Rust code for which we can retrieve wrappers.

yutannihilation:
Thanks, I'll use "example".

R/register_extendr.R
```diff
@@ -12,20 +12,45 @@
...
   pkg_name <- x$get("Package")
 
   if (!isTRUE(quiet)) {
     message(glue("Generating extendr wrapper functions for package: {pkg_name}"))
   }
 
+  entrypoint_c_file <- rprojroot::find_package_root_file("src", "entrypoint.c", path = ".")
+  if (!file.exists(entrypoint_c_file)) {
+    stop(
+      "src/entrypoint.c, which is necessary to make wrappers available to R sessions, doesn't exist.",
```
yutannihilation, line 32: Implemented the check on `src/entrypoint.c`. But I'm still thinking what the error message should be here...

R/register_extendr.R
```diff
@@ -12,20 +12,45 @@
...
+  # If FALSE, execute make_wrappers() only when the package can be loaded.
+  if (isTRUE(force_wrappers)) {
+    tryCatch(
+      make_wrappers(pkg_name, pkg_name, outfile, use_symbols = TRUE, quiet = quiet),
+      error = function(...) {
+        warning(
+          "Generating the wrapper functions failed, so a minimal one is used instead",
+          call. = FALSE
+        )
+        make_sample_wrappers(pkg_name, outfile)
```
clauswilke, line 50: I think this needs to be updated here also.

R/register_extendr.R
clauswilke: Maybe write: "Could not find `src/entrypoint.c`. Are you sure this package is using extendr Rust code?"

clauswilke:
I think you should also add yourself to the authors list in `DESCRIPTION`.

R/register_extendr.R
yutannihilation: Sounds good, thanks.

R/register_extendr.R
yutannihilation: Oh, good catch...

yutannihilation:
Merging now. It's an honor to join the authors list :)

Issue #54, opened 2021-03-06
Test `use_extendr()` on CI


Ilia-Kosenkov:
In light of #53, I suggest adding a separate step that creates an empty package and adds extendr dependencies.
If we add a simple test (`expect_equal(hello_world(), "Hello world!")`), then R CMD check will reveal any potential problems with extendr setup.

yutannihilation:
Sounds good. I think we can use `usethis::create_package()` to create an empty package.

https://usethis.r-lib.org/reference/create_package.html

Ilia-Kosenkov:
I have an example here
https://www.github.com/Ilia-Kosenkov/rextendr/tree/use-extendr-ci-test/.github%2Fworkflows%2FR-CMD-check.yaml

yutannihilation:
Thanks, I have some questions.

I think we should do this outside of the repository. 

``` r
d <- tempfile()
dir.create(d)
setwd(d)
```

Shouldn't this be done in `use_extendr()`?

``` r
usethis::use_build_ignore(".*\\.o$", escape = FALSE)
```

What's the intent of these two `devtools::document()`? I'm wondering if it should be like this instead:

``` r
devtools::install()
rextendr::register_extendr()
devtools::document()
```

Ilia-Kosenkov:
@yutannihilation,
I made a quick prototype. I am not sure about buildignore - `devtools::document` builds Rust source so package contains also build artifacts, like .o and .dll.

First `devtools::document` builds source and produces R code. Second call updates NAMESPACES and *rd docs.

If your solution is better, I have no objections.

yutannihilation:
>  I made a quick prototype. I am not sure about buildignore - `devtools::document` builds Rust source so package contains also build artifacts, like .o and .dll.

Ah, I see. Sorry, I was a bit confused.

> First `devtools::document` builds source and produces R code.

I mean, if I understand correctly, it doesn't. You need to call `rextendr::register_extendr()` explicitly to produce R code, and it needs the package to be installed to let `.Call()` find the symbol `wrap__make_<package>_wrappers`.

Ilia-Kosenkov:
@yutannihilation,
Then I am lost as this script works fine in my local environment. We need to test both approaches.

yutannihilation:
Yes, it works, but I think it's because `use_extendr()` creates `R/extendr-wrappers.R`, not that the compiled rust code.

(Sorry, I didn't follow the discussions around the wrapper generation, so I might be wrong...)

[...]

PR #61, opened 2021-03-08
Test package scaffolding on CI

Ilia-Kosenkov:
`use_extendr()` prepares new packages to work with extendr. It is useful to test if `use_extendr()` creates all the necessary files.
To do that, we can create an empty project, link it to extendr, and write a simple test. If R CMD check on this package succeeds, the setup was correct.

See #54 for a discussion on how to correctly build a package with `{rextendr}`.


clauswilke:
Somehow it's shown all sorts of changes that were already committed on the main branch. Did you do a merge commit with main? If not, I suggest you do one. That's always the easiest way to keep branches and main synchronized, and it'll disappear when we do the squash merge.

.github/workflows/R-CMD-check.yaml
```diff
@@ -102,3 +106,36 @@ jobs:
...
+              c(
+                  "test_that(\"`hello_world()` works\", {",
+                  "  expect_equal(hello_world(), \"Hello world!\")",
+                  "})"
+              ),
+              file.path("tests", "testthat", "test-hello.R")
+          )
+          rextendr::use_extendr()
+          devtools::document()
+          devtools::install()
```
clauswilke, line 132: I think you can use `devtools::install(quick = TRUE)` here.

Ilia-Kosenkov:
@clauswilke,
Oh sorry, I was in a hurry and did it wrong, I will redo it shortly.

.github/workflows/R-CMD-check.yaml
Ilia-Kosenkov: I fixed this.

clauswilke:
The latest version looks fine to me, but as is it doesn't test that `register_extendr()` actually updates the wrappers.

One way to do this, I think, is to test that the first line in the wrappers is `# Generated by extendr: Do not edit by hand`. This line gets added only after the wrappers are updated with `register_extendr()`, so you could test that it's absent after `use_extendr()` and present after `register_extendr()`.

Having said that, I'm a bit concerned about these massive github action scrips we're writing. Maybe move this test into a separate script, just so things get a little more manageable?

clauswilke (approved):
Thanks!

Ilia-Kosenkov:
@clauswilke,
Great, thank you, merging now.

Issue #59, opened 2021-03-07
rextendr::document()


yutannihilation:
(The original discussion: https://github.com/extendr/rextendr/issues/56#issuecomment-792186856)

Currently, we need to execute 3 lines of codes to reflect the changes on the Rust code. We might need some shortcut for this.

``` r
devtools::install(quick = TRUE)
rextendr::register_extendr()
devtools::document()
```

clauswilke:
The most elegant way to develop a shortcut is to call `roxygen2::roxygenise()` directly, with a custom load function that updates the wrappers and sources them after compiling and before documenting.

clauswilke:
If we copy the code from `devtools::document()` we run into a licensing issue as it's GPL licensed.

@jimhester @hadley Do we have permission to use the body of `devtools::document()` under MIT license? Specifically these lines: https://github.com/r-lib/devtools/blob/4a55464dee9cb99b97f7e9bb6c098e533ab1761f/R/document.R#L13-L35

clauswilke:
In theory, something like the following should work. In practice, though, I can't get this to use the newly compiled dll on macOS. It looks like the same problem we had previously with `rust_source()`.

```r
load_rextendr <- function(path) {
  pkgbuild::clean_dll(path = path)
  env <- pkgload::load_all(path, helpers = FALSE, attach_testthat = FALSE)$env
  eval(bquote(rextendr::register_extendr(.(path))), envir = env)
  wrappers <- rprojroot::find_package_root_file("R", "extendr-wrappers.R", path = path)
  source(wrappers, local = env)
  env
}


document <- function(pkg = ".", roclets = NULL, quiet = FALSE) {
  pkg <- devtools::as.package(pkg)
  if (!isTRUE(quiet)) {
    cli::cli_alert_info("Updating {.pkg {pkg$package}} documentation")
  }
  
  devtools:::save_all() # not currently exported

  if (quiet) {
    output <- tempfile()
    on.exit(unlink(output))
    withr::local_output_sink(output)
  }
  withr::local_envvar(devtools::r_env_vars())
  
  roxygen2::roxygenise(pkg$path, roclets, load_code = load_rextendr)
  pkgload::dev_topic_index_reset(pkg$package)
  invisible()
}
```

hadley:
Yes, that's fine — I think you can consider the majority of devtools to be MIT licensed, we just haven't rewritten the one function we copied from base R yet, so we can't consider the whole package to be MIT.

[...]

PR #63, opened 2021-03-11
Generate wrappers on compile

### LINKED ISSUE #56: Can we generate extendr-wrappers.R at the same time of compiling Rust code?

yutannihilation:
Close #56 

At the moment, this is just a very quick implementation, so I guess there's a lot to fix. I need your feedback!

After this gets merged, the conventions rextendr assumes will be:

1. `src/extendr-wrappers.R` exists, or will be generated on compile (mandatory)
2. `Makevars` has target `extendr-wrappers.R` (optional)

While `use_extendr()` generates `bin/generate_wrappers.rs`, users can specify arbitrary code to generate `src/extendr-wrappers.R`. `register_extendr()` calls `system2("make", c("-C", src_dir, "-f", "Makevars", "extendr-wrappers.R"))` to ensure the wrapper file is up-to-date. `Makevars` might not contain the target, so the failure is ignored and the existing `src/extendr-wrappers.R` will be used then.

When `src/extendr-wrappers.R` doesn't exist, it falls back to the current method of calling shared library. I'm wondering if this should be left or not.

Note that I felt `use_extendr()` is a bit lengthy, so I split into separate functions for each file. If this makes it difficult to review this, I'll revert it and do another PR.


yutannihilation:
Hmm, sorry, it seems I misunderstood how `cargo run` fails on Windows. I thought it's only on 32bit, but we cannot use `cargo run` at all?

Ilia-Kosenkov:
@yutannihilation,
It seems you do not specify `target`. On windows think of this as a mandatory parameter.
We might also think about updating makevars to have configurable tool chain on windows (via environment variable if it can be done).

yutannihilation:
Oops... Thanks for catching!

Ilia-Kosenkov:
@yutannihilation,
I am not sure what is wrong, I will check it locally. In the meantime, I think we should also specify `--release`. I think the absence of this flag triggers recompilation of all dependencies.

yutannihilation:
> I think the absence of this flag triggers recompilation of all dependencies.

Thanks, the reason I didn't add `--release` was that I thought debug build would be faster, but didn't notice this...

yutannihilation:
@Ilia-Kosenkov Sorry, this was my mistake. As described, `register_extendr()` runs `make`, but I used `Makevars` even on Windows...

Ilia-Kosenkov:
@yutannihilation,
It's OK, I am running your latest commit locally and it seems to work so far.
I do not understand why CI fails though.

yutannihilation:
Oh, it still fails. Hmm...

yutannihilation:
It seems I'm really good at making fatal typos...

Ilia-Kosenkov:
@yutannihilation,
Sooo, I am a little bit confused here. Can we now simplify this code?
https://github.com/extendr/rextendr/blob/429d21f07674c6c878ce1fda3d300adeae762af6/.github/workflows/test_pkg_gen.yaml#L117-L121
Can it be reduced to just `devtools::document()`?

yutannihilation:
Good point. In my understanding, we can omit the first `devtools::document()` and `devtools::install()`, but we still need to

``` r
rextendr::use_extendr()

# This is needed because compile doesn't automatically fish the generated
# src/extendr-wrappers.R to R/extendr-wrappers.
rextendr::register_extendr()

devtools::document()
```

Ilia-Kosenkov:
@yutannihilation,
Hm, while I am ok with the implementation, it does not significantly simplify the package development process, at least compared to my experiment with roxygen2/roclets.
In both cases we rely on something unstable like external `make` here and package loading/unloading in my case.
I, unfortunately, have no access to MacOS-based device to observe the problem @clauswilke mentioned, but perhaps I can reproduce it in CI.
The roclet solution allows removing all the extra boilerplate, leaving olnly
```r
devtools::document()
devtools::document()
```

Could you modify the CI to test how your solution works?

yutannihilation:
I don't understand how roclets works, but I vaguely thought it can call `rextendr::register_extendr()` inside so that we need to call `devtools::document()` only once.

> Could you modify the CI to test how your solution works?

Sure.

Ilia-Kosenkov:
@yutannihilation,
You are correct about roclets, but roclets code processing is immutable -- roxygen2 parses source and extracts all tags, then it feeds the same input to all roclets sequentially, so a roclet cannot depend on the output of another roclet. When our custom roclet generates new wrappers, the built-in roclet that writes the `NAMESAPCE` file does not see this changes.

I also found an issue with 

> ```r
> rextendr::use_extendr()
> 
> # This is needed because compile doesn't automatically fish the generated
> # src/extendr-wrappers.R to R/extendr-wrappers.
> rextendr::register_extendr()
> 
> devtools::document()
> ```

When we call `register_extendr()` first, there is no library built. So your `fish_wrappers()` initiates the first build, but it does so by calling simple `make`, which does not setup R-specific variables. In my local setup, this compilation failed because `make` was unable to correctly construct `$TARGET`.

yutannihilation:
I too was staring at that line. Yeah, it seems I was too optimistic about how complex to mimic the compilation by R properly...

yutannihilation:
Okay, let me withdraw the idea of invoking only the wrapper generation from `rextendr::register_extendr()`... I think I can pass the test by providing some variables manually in this simple case, but apparently users might write more complex `Makevars`/`Makevars.win` that my expectation cannot cover.

Ilia-Kosenkov:
@yutannihilation,
I think it is not complicated, R has some pre-defined makefiles stored in its folder which get included into the compilation process when it is down correctly. The question here is what is the best approach.
I am gonna move to #59 to summarize what I found is needed to make `*::document()` work as smoothly as possible.

yutannihilation:
I see.

clauswilke:
Btw., the `make` problems mostly disappear if you call `pkgbuild::compile_dll()`. I still don't think that's such a terrible idea.

yutannihilation:
Yeah, I don't give up this pull request yet. But, if loading/unloading is the only problem, I think callr will solve it. So, now I'm not sure which is better than the other.

clauswilke:
> But, if loading/unloading is the only problem, I think callr will solve it. So, now I'm not sure which is better than the other.

I think `callr()` is the better approach then. We want to put as few constraints on our downstream users as possible. Asking them to add weird code to `Makevars` that they don't understand and that looks confusing is unnecessary cognitive load. It's better if we can say "Use this simple `Makevars` file and run this simple function to update routine registration." And ideally it'll just work with `devtools::document()`, but even `rextendr::document()` would be fine.

PR #64, opened 2021-03-11
Use callr to solve load/unload problem

yutannihilation:
Using a separate R process should solve the problem.

### DIFF
--- a/R/register_extendr.R
+++ b/R/register_extendr.R
@@ -63,11 +63,17 @@ register_extendr <- function(path = ".", quiet = FALSE, force_wrappers = FALSE) 
 make_wrappers <- function(module_name, package_name, outfile,
                           use_symbols = FALSE, quiet = FALSE) {
   wrapper_function <- glue("wrap__make_{module_name}_wrappers")
-  x <- .Call(
-    wrapper_function,
-    use_symbols = use_symbols,
-    package_name = package_name,
-    PACKAGE = package_name
+  x <- callr::r_safe(
+    function(PACKAGE, ...) {
+      library(PACKAGE, character.only = TRUE)
+      .Call(..., PACKAGE = PACKAGE)
+    },
+    args = list(
+      wrapper_function,
+      use_symbols = use_symbols,
+      package_name = package_name,
+      PACKAGE = package_name
+    )
   )
   x <- stringi::stri_split_lines1(x)
 
--- a/R/register_extendr.R
+++ b/R/register_extendr.R
@@ -36,29 +36,25 @@ register_extendr <- function(path = ".", quiet = FALSE, force_wrappers = FALSE) 
 
   outfile <- rprojroot::find_package_root_file("R", "extendr-wrappers.R", path = path)
 
-  # If force_wrappers is TRUE, use tryCatch() to generate minimal wrappers even
-  # when there's some error (e.g. the symbol cannot be found).
-  # If FALSE, execute make_wrappers() only when the package can be loaded.
+  # If force_wrappers is TRUE, generate minimal wrappers even when there's some
+  # error (e.g. the symbol cannot be found).
   if (isTRUE(force_wrappers)) {
     error_handle <- function(e) {
-      warning(
-        "Generating the wrapper functions failed, so a minimal one is used instead",
-        call. = FALSE
-      )
+      msg <- "Generating the wrapper functions failed, so a minimal one is used instead"
+      warning(msg, call. = FALSE)
       make_example_wrappers(pkg_name, outfile)
     }
   } else {
     error_handle <- function(e) {
-      stop(
-        glue("Package {pkg_name} cannot be loaded. No wrapper functions were generated."),
-        call. = FALSE
-      )
+      stop("Generating the wrapper functions failed", call. = FALSE)
     }
   }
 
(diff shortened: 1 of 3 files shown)


yutannihilation:
Hmm, this needs a bit more consideration about how to support both the usage within on same session (e.g. `rust_function()`) and that on a separate session (e.g. `register_extendr()`), but I believe this should work in principle.

clauswilke:
At this point, maybe just write two separate functions, one to make wrappers for `rust_source()` and one to make wrappers for a package?

yutannihilation:
Thanks, I thought I could write:

``` r
callr::r_safe(make_wrappers, args = args)
```

But, as there's no package namespace on the new process, so this fails.

Next, I wondered if this works,

```r
function(package_root, ...) {
  pkgload::load_all(package_root, quiet = TRUE)
  make_wrappers <- get("make_wrappers", asNamespace("rextendr"))
  make_wrappers(...)
}
```

but probably there's a corner case when rextendr is not installed yet or outdated. So I ended up defining the function within `make_wrappers` and switch between `callr::r_safe()` and `do.call()`.

I think this is ready for review now.

R/register_extendr.R
```diff
@@ -36,39 +36,54 @@ register_extendr <- function(path = ".", quiet = FALSE, force_wrappers = FALSE)
 
   outfile <- rprojroot::find_package_root_file("R", "extendr-wrappers.R", path = path)
 
-  # If force_wrappers is TRUE, use tryCatch() to generate minimal wrappers even
-  # when there's some error (e.g. the symbol cannot be found).
-  # If FALSE, execute make_wrappers() only when the package can be loaded.
+  # If force_wrappers is TRUE, generate minimal wrappers even when there's some
+  # error (e.g. the symbol cannot be found).
```
clauswilke, line 40: It's not clear to me here what "the symbol" is. Maybe rephrase?

R/register_extendr.R
```diff
@@ -36,39 +36,54 @@ register_extendr <- function(path = ".", quiet = FALSE, force_wrappers = FALSE)
...
-          "Generating the wrapper functions failed, so a minimal one is used instead",
-          call. = FALSE
-        )
-        make_example_wrappers(pkg_name, outfile)
-      }
-    )
-  } else if (requireNamespace(pkg_name, quietly = TRUE)) {
-    make_wrappers(pkg_name, pkg_name, outfile, use_symbols = TRUE, quiet = quiet)
+    error_handle <- function(e) {
+      msg <- "Generating the wrapper functions failed, so a minimal one is used instead"
```
clauswilke, line 43: The word "one" doesn't refer to anything in this sentence. (It can't refer to "wrapper functions", as that is plural.) Maybe write "a minimal wrapper file"?

R/register_extendr.R
```diff
@@ -36,39 +36,54 @@ register_extendr <- function(path = ".", quiet = FALSE, force_wrappers = FALSE)
...
 
 make_wrappers <- function(module_name, package_name, outfile,
-                          use_symbols = FALSE, quiet = FALSE) {
+                          use_symbols = FALSE, quiet = FALSE,
+                          use_callr = FALSE) {
   wrapper_function <- glue("wrap__make_{module_name}_wrappers")
-  x <- .Call(
+
+  func <- function(package_root, ...) {
+    pkgload::load_all(package_root, quiet = TRUE)
```
clauswilke, line 70: I don't think this line should be executed when we're running `rust_source()`. I'm still wondering whether it makes sense to try to let `make_wrappers()` handle two so very different use cases. It feels forced.

yutannihilation: Ah, true...

yutannihilation:
The solution was simple; we can just pass `make_wrappers` as an argument.

``` r
function(package_root, make_wrappers, ...) {
  pkgload::load_all(package_root, quiet = TRUE)
  make_wrappers(...)
}
```

I'll improve those cryptic comments and messages later... Thanks for catching.

clauswilke:
Thanks, this looks good. I think loading the package in a separate process is safer anyways, as it doesn't pollute the current global environment. I don't like `load_all()` when developing, because it's polluting my interactive session, so this is good regardless of whether it's strictly necessary for success or not.

clauswilke (approved):

yutannihilation:
Thanks. I chose `load_all()` not only because using a separate process is safe, but also because I'm afraid `devtools::install()` might be unsafe when there's multiple R processes. Especially, it's one of the common failures on Windows to fail to copy a DLL when it is open by another process (`problem copying foo.dll ...: Permission denied`).

clauswilke:
Yes, it has to be `load_all()`. I was just trying to say that out-of-process calling was probably needed even if reloading an existing dll was always possible without problems.

yutannihilation:
Ah, got it.

yutannihilation:
I'll hold this pull request until #67 gets merged. As this is a smaller change, I expect it's easier to fix conflicts here than there.

clauswilke:
I think we should merge this first. The change to `register_extendr()` is needed to make `document()` work. Can we merge this as is or does anything else need to be changed?

yutannihilation:
I see. Let's merge this.

PR #66, opened 2021-03-12
fix vignette

clauswilke:
I just noticed that a vignette had some outdated code examples.

### DIFF
--- a/vignettes/rmarkdown.Rmd
+++ b/vignettes/rmarkdown.Rmd
@@ -120,10 +120,10 @@ rprintln!("x = {}, y = {}", x, y);
 
 ## Exporting Rust functions to R
 
-The chunk type `extendrfuns` compiles Rust functions and registers them with R to be called later. For example, consider this code chunk. It creates R functions `hello()` and `foo()`.
+The chunk type `extendrsrc` compiles Rust functions and objects and registers them with R to be called later. For example, consider this code chunk. It creates R functions `hello()` and `foo()`, as well as a class called `Counter`.
 
 ````markdown
-`r ''````{extendrfuns}
+`r ''````{extendrsrc}
 #[extendr]
 fn hello() -> &'static str {
     "Hello from Rust!"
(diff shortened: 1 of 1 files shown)


vignettes/rmarkdown.Rmd
```diff
@@ -120,10 +120,10 @@ rprintln!("x = {}, y = {}", x, y);
 
 ## Exporting Rust functions to R
```
yutannihilation, line 121: Should this also be changed to "functions and objects"?

yutannihilation (commented):
Thanks for catching. This sentence should also be fixed

https://github.com/extendr/rextendr/blob/2c27093af86ddba968e0717295c4600142cb16b8/vignettes/rmarkdown.Rmd#L15

vignettes/rmarkdown.Rmd
clauswilke: Yes, probably, thanks!

clauswilke:
@yutannihilation Can you look it over one more time. I made a few more changes.

vignettes/rmarkdown.Rmd
```diff
@@ -64,12 +64,19 @@ x <- 5.3
 y <- "hello"
 ```
 
-We can read these variables from Rust using the `R!()` macro. Unfortunately, it takes a bit of unwrapping and type-converting to get these into native Rust values. This is the case because things could go wrong at each step, and we're ignoring the errors with `unwrap()`. For example, `R!()` calls the R interpreter and it could generate an error. And even if it returns a value, we're not guaranteed that the value can be coerced into a real value or a string, so `as_real()` and `as_str()` can also error out.
+We can read these variables from Rust using the `R!()` macro. Unfortunately, it takes a bit of error checking and type-converting to get these into native Rust values. This is the case because things could go wrong at each step. We're handling the errors with `?`. For example, `R!()` calls the R interpreter and it could generate an error. And even if it returns a value, we're not guaranteed that the value can be coerced into a real value or a string. However, `as_real()` and `as_str()` don't return errors, they return options, so we need to turn the `None` option into an error via `ok_or()`.
```
yutannihilation, line 67: Let me clarify. Do we really want to explain how to handle errors? I think error handling is a bit advanced topic (e.g. how to return `NA` or a default value when it fails) that we cannot cover briefly here, so it might be valid to assume only the success case and unwrap them naively.

clauswilke: It bothers me that most Rust demonstration code uses `unwrap()` and then they tell you to never actually use `unwrap()` in production. I'd rather write demo code that is suitable for production.

Whether the error handling should be explained is a separate question. I can shorten the text if you think that would be helpful.

yutannihilation: Hm, my feeling is that when we write `extendr` chunk I bet it's for an illustration purpose in most of the cases so `unwrap()` should be fine. That said, I get your point. Let's go ahead.

yutannihilation (approved):

PR #65, opened 2021-03-11
Roxygen integration (wrapper generation using roclets)

Ilia-Kosenkov:
This continues the discussions from #56, #59, #63, #64.
In this PR we make use of roxygen2 roclet feature, effectively defining a custom script that is injected in the middle of `roxygen2::roxygenize()` process.

Primary modifications:
1. Improved `use_extendr` to write additional metadata necessary to enable roclets support
2. Implemented `registration_roclet` which simply calls `rextendr::register_extendr()`.
3. Implemented a roclet to track changes in Rust files, similar to what `pkgbuild::needs_compilation()` does.

Issue #62, opened 2021-03-08
Suggesting or replicating usethis functions


malcolmbarrett:
I'm happy to see the emergence of usethis-style functions, e.g. #53 and #60. 

I always find it a bit strange, though, when usethis-style functions don't quite behave like usethis (e.g. the UI and other defaults). A simple example is using `message()` instead of `ui_*()` or cli.

One solution is to add usethis to Suggests. Another option is to create lighter-weight internal options using cli. 

I'd be happy to work on this if it seems like a good idea.

clauswilke:
That's mostly just my ignorance. Feel free to change. I'd rather have a cli dependency than usethis dependency.

malcolmbarrett:
Ok, I'll prep a PR exploring that. Thanks!

Ilia-Kosenkov:
I just want to point out that here https://github.com/r-lib/usethis/issues/956 hadley recommends to use `{cli}` instead of `ui_*` (at least that's how I understand his message). I suggest to use `{cli}` (see some crude examples in #65).

malcolmbarrett:
Oh, great that you're already planning to import cli. Yes, there's some nice guidance here for writing usethis-like ui: https://cli.r-lib.org/articles/usethis-ui.html. I think limiting to cli makes sense for this use case.

Ilia-Kosenkov:
Please take into account this set of functions when adding {cli} support
https://github.com/extendr/rextendr/blob/main/R/ui.R

malcolmbarrett:
Thanks! I had intended to that in this PR, so awesome 😎 

I might continue waiting on this until #88/#70 are resolved as that directly affects the approach for this small issue

Ilia-Kosenkov:
@malcolmbarrett,
Should this be closed or are there any other features you think should be implemented?

malcolmbarrett:
I've got one more PR coming surrounding the way templates are handled that will close this

PR #67, opened 2021-03-12
rextendr::document() infrastructure for package development

Ilia-Kosenkov:
This PR implements `rextendr::document()`, which performs additional tasks prior to calling `devtools::document()`, solving most of the Rust compilation and wrapper generation problems. 
This is based on #65, though roclet support has been fully retired.

The new workflow is fairly simple:
```R
devtools::create(".")
rextendr::use_extendr()
rextendr::document()
hello_world()
```


R/rextendr_document.R
```diff
@@ -0,0 +1,23 @@
...
+#' @export
+document <- function(pkg = ".", quiet = FALSE, roclets = NULL) {
+  try_save_all()
+  needs_compilation <- needs_compilation(pkg, quiet) || pkgbuild::needs_compile(pkg)
+
+  pkgload::load_all(
+    path = pkg,
+    compile = needs_compilation,
+    quiet = quiet
+  )
```
yutannihilation, line 19: Note that we don't need this after #64 gets merged.

R/register_extendr.R
```diff
@@ -15,18 +15,22 @@
 #' @param force_wrappers Logical indicating whether to generate a minimal
 #'   wrapper in the cases when the package's namespace cannot be loaded. This is
 #'   useful to recover the wrapper file when something is wrong with it.
-#' @return The generated wrapper code. Note that this is not normally needed,
-#' as the function saves the wrapper code to `R/extendr-wrappers.R`.
+#' @return Logical scalar. `TRUE` if new wrappers were emitted, `FALSE`
+#'   if wrappers did not require an update.
```
yutannihilation, line 19: Isn't it more straightforward to show some message like "Skip generating wrappers"?

yutannihilation (commented):
Minor comments.

R/rextendr_document.R
clauswilke: That's the argument for me to merge #64 first.

yutannihilation:
@Ilia-Kosenkov 
I made a quick pull request to resolve conflicts introduced by merging #64. Hope this helps.

https://github.com/Ilia-Kosenkov/rextendr/pull/1

Ilia-Kosenkov:
There is some issue with the path handling. I am unable to run `rextendr::document('my_package')` from the package folder's parent. Need time to investigate.

R/register_extendr.R
```diff
@@ -72,27 +81,59 @@ make_wrappers <- function(module_name, package_name, outfile,
...
     make_wrappers = make_wrappers,
+    compile = compile,
+    # arguments passed to make_wrappers()
     module_name = module_name,
     package_name = package_name,
     outfile = outfile,
     use_symbols = use_symbols,
-    quiet = quiet
+    quiet = quiet,
+    path = path
```
yutannihilation, line 136: Sorry, probably this should be `rprojroot::find_package_root_file(path = path)` as well.

Ilia-Kosenkov: Yeah I found a few places, if it works locally I will push the updates.

R/register_extendr.R
Ilia-Kosenkov: My idea is that because it is an exported function that *may* be called separately at some point, it will provide information about the package state: `FALSE` if things are up to date, `TRUE` if new wrappers were generated, and exception if things are bad.

yutannihilation: Okay, makes sense. Then, how about returning the paths to the updated files (i.e. `R/extendr-wrappers.R`)? That's consistent with what the equivalents of cpp11 and Rcpp do:

- https://cpp11.r-lib.org/reference/cpp_register.html#value
- https://github.com/RcppCore/Rcpp/blob/c5a97a3092bb038dc4236133144ebdb877764a72/man/compileAttributes.Rd#L29

Ilia-Kosenkov:
In my original concept, I separated package compilation and wrapper generation into different tasks. With the current approach, everything gets piled up in `make_wrappers_externally`. As a result, `needs_new_wrappers()` no longer works correctly. This requires some overhaul. I will hopefully fix it today.

R/track_rust_source.R
```diff
@@ -0,0 +1,110 @@
...
+    return(FALSE)
+  }
+
+  # Obtains detailed info of each source file and library file.
+  # This includes 'last modification time'.
+  # Stored as tibbles
+  rust_info <- fs::file_info(rust_files)
+  library_info <- fs::file_info(library_path)
+
+  # Leaves files, which were modified *after* the library
```
clauswilke, line 86: Minor point: Write "Leave files that were modified" It's a restrictive clause.

See: https://www.lavc.edu/getattachment/writingcenter/Handouts/Clauses-III_-Restrictive_Nonrestrictive-ADA-2018.pdf

R/track_rust_source.R
```diff
@@ -0,0 +1,110 @@
...
+
+#' Checks if re-compilation is needed.
+#' Tracks changes in Rust source files (`*.rs`) and `Cargo.toml`.
+#'
+#' @param path Package root.
+#' @param quiet Logical scalar indicating wether the output should be quiet (`TRUE`)
+#'   or verbose (`FALSE`).
+#' @returns Logical `TRUE` if Rust source has been modified, `FALSE` otherwise.
+#' @export
+needs_compilation <- function(path = ".", quiet = FALSE) {
```
clauswilke, line 60: Do we need to export this function? At a minimum, add `@keywords internal` so it gets excluded from the documentation index.

R/use_extendr.R
```diff
@@ -55,10 +51,10 @@ void R_init_{pkg_name}(void *dll) {{
 }}
 )"
   )
-  brio::write_lines(entrypoint_content, file.path(src_dir, "entrypoint.c"))
+  usethis::write_over(fs::path(src_dir, "entrypoint.c"), entrypoint_content, quiet = quiet)
```
clauswilke, line 54: Is there any good reason to use `usethis::write_over()` rather than `brio::write_lines()`. It's the main reason we need the {usethis} dependency, I believe.

R/use_extendr.R
```diff
@@ -152,16 +146,19 @@ extendr_module! {{
...
+    cli::cli_alert_success("Finished configuring {.pkg extendr} for package {.pkg {pkg_name}}.")
+    cli::cli_alert_warning("Please update the system requirement in {.file DESCRIPTION} file.")
+    cli::cli_alert_warning("Please run {.fun rextendr::document} for changes to take effect.")
   }
 
   return(invisible(TRUE))
 }
 
-make_example_wrappers <- function(pkg_name, outfile, extra_items = NULL) {
+make_example_wrappers <- function(pkg_name, outfile, extra_items = NULL, quiet = FALSE, path = ".") {
```
clauswilke, line 161: I would move the path argument further up, certainly before `quiet` and maybe also before `extra_items`. Also, it would be good to add some comments explaining what the function and the various arguments do. For internal functions, I often like to mimic roxygen2 without actually using it, e.g.:

```r
# Function to generate a simple wrapper file if no autogenerated wrappers are available yet
#
# @param pkg_name ...
# @param outfile ...
```
```

File: R/use_extendr.R [hunk 8]
```diff
@@ -181,4 +178,20 @@ make_example_wrappers <- function(pkg_name, outfile, extra_items = NULL) {
   }
 
   brio::write_lines(wrappers_content, outfile)
+  if (!isTRUE(quiet)) {
+    rel_path <- pretty_rel_path(outfile, search_from = path)
+    cli::cli_alert_success("Writting wrappers to {.file {rel_path}}.")
+  }
+}
+
+write_namespace <- function(pkg_name, path = ".", quiet = FALSE) {
```
clauswilke, line 187: I would prefer if we didn't have this function at all. It feels like mission creep to me.

R/use_extendr.R
```diff
@@ -152,16 +146,19 @@ extendr_module! {{
     )"
   )
 
-  make_example_wrappers(pkg_name, wrappers_file, extra_items = example_function_wrapper)
+  make_example_wrappers(pkg_name, wrappers_file, extra_items = example_function_wrapper, path = path)
+  write_namespace(pkg_name, path, quiet = quiet)
```
clauswilke, line 150: I don't like that we're modifying the namespace file. In particular since we're then telling the user to run `rextendr::document` for the changes to take effect. Wouldn't that call generate the namespace file? Is there something I'm missing? In either case, let's leave modifying the namespace file to `devtools::document()`. We can call that if necessary.

R/track_rust_source.R
```diff
@@ -0,0 +1,110 @@
...
+  fs::path(
+    glue::glue(
+      rprojroot::find_package_root_file("src", path = path),
+      "{pkg_name}{.Platform$dynlib.ext}",
+      .sep = .Platform$file.sep
+    )
+  )
+}
+
+get_rust_files <- function(path = ".") {
```
clauswilke, line 22: Please add a few comments explaining what this function does.

R/try_save_all.R
```diff
@@ -0,0 +1,6 @@
+try_save_all <- function() {
```
clauswilke, line 1: Please add a few comments explaining what this function does.

clauswilke (commented):
Thanks! Overall looks very good. I have a few comments throughout. Most importantly, we should get into the habit of documenting unexported helper functions.

There's also one remaining workflow issue. I don't like that we're modifying the namespace file directly, and I'm not sure we need to. I realize that the first time we run `rextendr::document()` after `rextendr::use_extendr()` we won't get the wrappers from the Rust library, but I think that's fine since `rextendr::use_extendr()` can just generate wrappers that are consistent with the Rust library it produces.

DESCRIPTION
```diff
@@ -45,11 +45,18 @@ Imports:
     tibble,
     purrr,
     rlang,
-    rprojroot
+    rprojroot,
+    usethis,
+    fs,
```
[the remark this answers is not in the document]
Ilia-Kosenkov: The only reason I used {fs} is that it provided an easy way to 'touch' files, `fs::file_touch`, which was used to modify `Makevars`, triggering recompilation. This is no longer needed, so {fs} can be safely removed. I suggest doing it at a later phase when other parts of `rextendr::document()` are finalized & tested.

{usethis} is a more complicated topic. There are definitely some useful functions, but we can have simple replacements. If we position {rextendr} as {usethis} for extendr, perhaps having {usethis}-like behavior and {usethis} as dependency is not that bad. It is not a problem to remove it though, so it is up to you.

R/track_rust_source.R
```diff
@@ -0,0 +1,110 @@
...
+
+get_library_path <- function(path = ".") {
+  # Constructs path to the library file (e.g., package_name.dll)
+  pkg <- desc::desc(rprojroot::find_package_root_file("DESCRIPTION", path = path))
+  pkg_name <- pkg$get("Package")
+  fs::path(
+    glue::glue(
+      rprojroot::find_package_root_file("src", path = path),
+      "{pkg_name}{.Platform$dynlib.ext}",
+      .sep = .Platform$file.sep
```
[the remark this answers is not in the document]
Ilia-Kosenkov: I will fix this.

R/track_rust_source.R
Ilia-Kosenkov: That depends. My reasoning was `pkgbuild::needs_compile()` is exported, so perhaps we can provide a similar API for Rust code. The function has no side effects so can be safely used by a package developer. I can easily remove `@export`.

R/use_extendr.R
Ilia-Kosenkov: There are several key differences between `brio::write_lines()` and `usethis::write_over()`:
1. `write_over()` informs the user if such file already exists. The user is asked whether to replace the file or not.
I find it useful in `use_extendr`, where we write a bunch of files. Imagine that the user already created `Makevars`, and we silently overwrite it. `write_over()` can help us avoid this.
2. `write_over()` output nice, styled message, something like `Writting to {file}`, which can be suppressed using `quiet` argument. with `write_lines()` we have to have an additional `if (!isTRUE(quiet)) { cli::cli_alert_success("Writing to {.file path}")}`
3. `write_over()` is useless for wrapper generation, because we need to have the most recent wrappers and can safely overwrite the existing file.

If {usethis} stays as package dependency, I think `write_over()` can be used for package set up. If not, we can either overwrite everything (not sure this is a great user experience) or make our own function (again, a few lines of code, nothing fancy or complicated).

R/use_extendr.R
Ilia-Kosenkov: Ok, will do.

R/use_extendr.R
Ilia-Kosenkov: Yeah, I agree. I did not actually check if we still need namespace modification. Before there were problems during the first `devtools::document()` -- during first package loading, `NAMESPACE` was missing `useDynLib`, and the native library was not linked, breaking all `.Call`s into it, including wrapper generation. With manually writing `useDynLib` into the `NAMESPACE`, first package loading also loads freshly compiled library, and all `.Call`s are OK. 

I'll let you know if we still need this.

R/use_extendr.R
Ilia-Kosenkov: As I said above, before `rextendr::document`, it was helping with the first-time wrapper generation.

Ilia-Kosenkov:
I do not like how we handle exceptions thrown by out-of-process wrapper generation
https://github.com/Ilia-Kosenkov/rextendr/blob/9cd8b5f30838d18b2ccc0faee490ec627b442b53/R/register_extendr.R#L45-L59
Earlier today I was debugging a path issue in `make_wrappers_externally`, and I was getting almost no information out of the failing build process. 

1. We need to use `e` and extract its `$message`, displaying it to the user. 
2. I enabled output from `callr::r()` using the value of `quiet` arg. I believe the compilation process should be visible, otherwise it looks like the R session is frozen (e.g., on one of my laptops the first-time compilation of Rust code takes up to a minute and I was staring at the blank terminal not understanding what is going on under the hood).

While `usethis::ui_*` has capabilities for throwing errors, `cli` lacks it. I am not sure if we are keeping `usethis`, so I'd like to suggest the following:
1. Introduce internal functions like `ui_x <- function(text)`, which should generate an ansi string with a red `x` mark at the beginning, e.g.:
```R
ui_x("Unable to compile library)
# > x Unable to compile library
```
We can also have `ui_i` and `ui_v` for info and success (any other names would do).
2. An error generating function, something like 
```R
ui_throw <- function(message, details)
```
Here `message` is a required string scalar, the primary error message, and details is a character vector of additional information, potentially formatted using `ui_*` functions.
The error reporting would then look like this:

```R
if (something_bad$has_happened()) {
  ui_throw(
    "Something bad has happened!",
     c(
      ui_x("This thing happened"),
      ui_x("That thing happened"),
      ui_i("Do this to avoid problems in future")
     )
}
```
Which produces the following error (thrown by `stop()`):
```
Error: Something bad has happened!
x This thing happened
x That thing happened
i Do this to avoid problems in future
```
Obviously, the output will be styled using `cli` and thus have fancy &  colored symbols on supported platforms.

I am unaware of any other easy-to-use approach that will allow us to throw errors that interrupt function executrion.
I am open to suggestions. 
This will also save us from code repetition whenever we throw an error.
I also understand this may conflict with #62 and @malcolmbarrett, so perhaps we should somehow cooperate if my sugegstion gets approved.

R/register_extendr.R
```diff
@@ -72,27 +88,62 @@ make_wrappers <- function(module_name, package_name, outfile,
...
-    message("Writing wrappers to:\n", outfile)
+    rel_path <- pretty_rel_path(outfile, search_from = path)
+    cli::cli_alert_success("Writting wrappers to {.file {rel_path}}.")
   }
-  brio::write_lines(x, outfile)
+}
+
+# TODO: This no longer works (and perhaps no longer needed).
+# Checks if new wrappers should be generated
+needs_new_warppers <- function(path = ".", wrapper_path = fs::path("R", "extendr-wrappers.R")) {
```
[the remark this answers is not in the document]
Ilia-Kosenkov: This has been removed.

R/register_extendr.R
```diff
@@ -72,27 +88,62 @@ make_wrappers <- function(module_name, package_name, outfile,
...
+  if (!fs::file_exists(library_path)) {
+    # No library found. This means this is likely the first run
+    # and wrappers are needed. This will trigger recompilation.
+    return(TRUE)
+  }
+
+  wrapper_info <- fs::file_info(wrapper_path)
+  library_info <- fs::file_info(library_path)
+
+  # If wrapeprs are older than the library file, new wrappers are needed.
```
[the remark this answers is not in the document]
Ilia-Kosenkov: I think this has also been removed.

R/use_extendr.R
Ilia-Kosenkov: I am afraid we need to make sure `NAMESPACE` contains `useDynLib`, otherwise initial compilation will fail.
Here is a reprex whcih does not modify `NAMESPACE`:
<details>
<summary> Reprex </summary>

``` r
temp_dir <- tempdir()
pkg_dir <- file.path(temp_dir, "testpkg")
dir.create(pkg_dir, recursive = TRUE)
setwd(pkg_dir)

devtools::create(".")
#> v Setting active project to 'C:/Users/[redacted]/AppData/Local/Temp/RtmpeKV5JY/testpkg'
#> v Creating 'R/'
#> v Writing 'DESCRIPTION'
#> Package: testpkg
#> Title: What the Package Does (One Line, Title Case)
#> Version: 0.0.0.9000
#> Authors@R (parsed):
#>     * First Last <first.last@example.com> [aut, cre] (YOUR-ORCID-ID)
#> Description: What the package does (one paragraph).
#> License: `use_mit_license()`, `use_gpl3_license()` or friends to pick a
#>     license
#> Encoding: UTF-8
#> LazyData: true
#> Roxygen: list(markdown = TRUE)
#> RoxygenNote: 7.1.1
#> v Writing 'NAMESPACE'
#> v Setting active project to '<no active project>'
rextendr::use_extendr()
#> v Creating src/rust/src.
#> v Writing 'C:/Users/[redacted]/AppData/Local/Temp/RtmpeKV5JY/testpkg/src/entrypoint.c'
#> v Setting active project to '<no active project>'
#> v Writing 'C:/Users/[redacted]/AppData/Local/Temp/RtmpeKV5JY/testpkg/src/Makevars'
#> v Setting active project to '<no active project>'
#> v Writing 'C:/Users/[redacted]/AppData/Local/Temp/RtmpeKV5JY/testpkg/src/Makevars.win'
#> v Setting active project to '<no active project>'
#> v Writing 'C:/Users/[redacted]/AppData/Local/Temp/RtmpeKV5JY/testpkg/src/.gitignore'
#> v Setting active project to '<no active project>'
#> v Writing 'C:/Users/[redacted]/AppData/Local/Temp/RtmpeKV5JY/testpkg/src/rust/Cargo.toml'
#> v Setting active project to '<no active project>'
#> v Writing 'C:/Users/[redacted]/AppData/Local/Temp/RtmpeKV5JY/testpkg/src/rust/src/lib.rs'
#> v Setting active project to '<no active project>'
#> v Writting wrappers to R/extendr-wrappers.R.
#> v Finished configuring extendr for package testpkg.
#> ! Please update the system requirement in DESCRIPTION file.
#> ! Please run `rextendr::document()` for changes to take effect.
rextendr::document()
#> ! No library found at src/testpkg.dll, recompilation is required.
#> i Generating extendr wrapper functions for package: testpkg.
#> i Loading testpkg
#> Re-compiling testpkg
#>   -  installing *source* package 'testpkg' ...
#>      ** using staged installation
#>      ** libs
#>      rm -Rf testpkg.dll ./rust/target/x86_64-pc-windows-gnu/release/libtestpkg.a entrypoint.o
#>      "C:/rtools40/mingw64/bin/"gcc  -I"C:/PROGRA~1/R/R-40~1.4/include" -DNDEBUG          -O2 -Wall  -std=gnu99 -mfpmath=sse -msse2 -mstackrealign -UNDEBUG -Wall -pedantic -g -O0 -c entrypoint.c -o entrypoint.o
#>      cargo build --target=x86_64-pc-windows-gnu --lib --release --manifest-path=./rust/Cargo.toml
#>              Updating crates.io index
#>             Compiling winapi-build v0.1.1
#>       Compiling winapi-x86_64-pc-windows-gnu v0.4.0
#>       Compiling winapi v0.3.9
#>       Compiling winapi v0.2.8
#>       Compiling proc-macro2 v1.0.24
#>       Compiling unicode-xid v0.2.1
#>       Compiling syn v1.0.63
#>             Compiling extendr-engine v0.2.0
#>             Compiling lazy_static v1.4.0
#>             Compiling kernel32-sys v0.2.2
#>             Compiling quote v1.0.9
#>             Compiling extendr-macros v0.2.0
#>             Compiling libR-sys v0.2.1
#>             Compiling extendr-api v0.2.0
#>             Compiling testpkg v0.1.0 (C:\Users\i[redacted]\AppData\Local\Temp\RtmpeKV5JY\testpkg\src\rust)
#>              Finished release [optimized] target(s) in 36.67s
#>      C:/rtools40/mingw64/bin/gcc -shared -s -static-libgcc -o testpkg.dll tmp.def entrypoint.o -L./rust/target/x86_64-pc-windows-gnu/release -ltestpkg -lws2_32 -ladvapi32 -luserenv -LC:/PROGRA~1/R/R-40~1.4/bin/x64 -lR
#>      installing to C:/Users/[redacted]/AppData/Local/Temp/Rtmpy8Equp/devtools_install_5df82b6e6ce7/00LOCK-testpkg/00new/testpkg/libs/x64
#>   -  DONE (testpkg)
#> Error in .Call(wrapper_function, use_symbols = use_symbols, package_name = package_name,  : 
#>   "wrap__make_testpkg_wrappers" not available for .Call() for package "testpkg"
#> Error: Failed to generate wrapper functions
```

<sup>Created on 2021-03-13 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

</details>

DESCRIPTION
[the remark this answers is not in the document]
clauswilke: My thinking is that maybe at some point in the future {usethis} will provide {rextendr} functionality by calling {rextendr} functions. I don't like circular dependencies, so I'd rather not have {rextendr} depend on {usethis}.

R/use_extendr.R
clauswilke: I feel this is a problem in the function that is making the wrappers. It should silently (or with a warning) ignore the fact that we don't have a wrapper generating function, instead of erroring out.

DESCRIPTION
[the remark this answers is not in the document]
Ilia-Kosenkov: I understand this, we can certainly get rid of {usehis}. 
I have much less experience in serious package development, but this idea seems counter-intuitive to me.
Ideally, `rextendr::use_extendr` should be pulled into {usethis}, it has no dependencies and simply writes a couple of files. In this case, it is perfectly reasonable to communicate with the user using {usethis} functions, specifically designed for that. 
If this happens, then `rextendr::use_extendr` can call `usethis::use_extendr`, as pretty much a lot of similar packages do.
Basically, I am imagining a sort of inverted dependency here. 
In addition, you suggest to ask {usethis} maintainers to introduce an extra dependency on {rextendr} just to call one function, which I doubt they would like. Not every {usethis} consumer will use extendr, but most of {rextendr} consumers will likely use {usethis} to set up and develop their packages, so it feels more natural for {rextendr} to depend on {usethis} (and make use of it), and not vice verse.

R/use_extendr.R
Ilia-Kosenkov: I may be wrong here, but without `NAMESPACE` entry we will not be able to compile code and generate wrappers in one call to `rextendr::document()`. Even if we somehow bypass this, by the end of the first compilation no exported function would work: even though the library is compiled, it is not dynamically loaded when the package is loaded (because of the missing `useDynLoad`). This is exactly what happens when we call `pkgload::load_all` and then attempt to get wrappers -- the library is compiled, the package is loaded, but without the library. No `.Call`s work. If we ensure that `NAMESPACE` contains `useDynLib`, we will be able to make `.Call`s and generate wrappers. This is what currently happens.

Basically, we write to `NAMESPACE` before it gets updated by {roxygen2}. After the first `rextendr::document()` the `NAMESPACE` gets rewritten and we get a `useDynLib` entry because we have an attribute in wrappers. After this, we never ever touch `NAMESPACE` again.

I have no better solution and am open to suggestions.

R/track_rust_source.R
Ilia-Kosenkov: Thank you for the useful link!

R/register_extendr.R
Ilia-Kosenkov: Done.

R/use_extendr.R
clauswilke: > Imagine that the user already created `Makevars`

I believe in my first draft of all of this I wasn't doing anything if `src` was already in place. Is this check still there or has it been removed?

R/use_extendr.R
clauswilke: I was thinking the following, starting from an empty project

1. `rextendr::use_extendr()`: Basic wrapper file gets written
2. `rextendr::document()` (or alternatively, `devtools::document()`, it's the same here): `NAMESPACE` gets updated, this transfers `useDynLoad` into `NAMESPACE`
3. `devtools::install()`: dynamic library gets built for the first time

clauswilke:
We had a conversation about {fs} somewhere but I can't find it anymore. In any case, it may not be worth it to go through the effort of removing this dependency. Let's leave it in.

R/use_extendr.R
yutannihilation: FWIW, with cpp11, it's a user's job to ensure `@useDynLib` is in `NAMAESPACE` (via the package doc). So, I'm wondering if we are doing too much.

```
> usethis::use_cpp11()
✓ Creating 'src/'
✓ Adding '*.o', '*.so', '*.dll' to 'src/.gitignore'
● Copy and paste the following lines into 'R/<package name>-package.R':
  ## usethis namespace: start
  #' @useDynLib <package name>, .registration = TRUE
  ## usethis namespace: end
  NULL
  [Copied to clipboard]
✓ Adding 'cpp11' to LinkingTo field in DESCRIPTION
✓ Adding 'C++11' to SystemRequirements field in DESCRIPTION
✓ Writing 'src/code.cpp'
● Modify 'src/code.cpp'
```

To do this programmatically, we need to tweak `NAMESPACE` anyway, though.

https://cpp11.r-lib.org/reference/cpp_register.html#examples

yutannihilation:
> I enabled output from `callr::r()` using the value of `quiet` arg. I believe the compilation process should be visible, otherwise it looks like the R session is frozen (e.g., on one of my laptops the first-time compilation of Rust code takes up to a minute and I was staring at the blank terminal not understanding what is going on under the hood).

Sorry, I didn't notice `show` arg... Thanks for finding it!

DESCRIPTION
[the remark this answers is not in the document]
yutannihilation: I have no strong opinion here, but I feel it's fair to have usethis as a dependency. Since we follow what usethis does and assumes the user is already familiar with usethis-like workflow, rextendr shares its fate with usethis to some extent.

R/use_extendr.R
clauswilke: Yes, I'm aware, and originally I didn't like that extendr writes `@useDynLib` into the wrappers, but now I think it's probably the right approach. I'm not convinced that the cpp way is done for some fundamental reasons based in first principles. It's most likely historic, that Rcpp was written before people even were widely using roxygen2. If you started from scratch, and you had some autogenerated code, wouldn't you want that code to include all the boilerplate that is needed?

In the end, if it comes down to depending on usethis for a convenient user experience, I'm fine with going that route. I just want to make sure we are somewhat careful in the dependencies we add. But any dependency that adds real value is fine by me.

DESCRIPTION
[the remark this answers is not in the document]
clauswilke: See my other comment here: https://github.com/extendr/rextendr/pull/67#discussion_r593830791

In brief, I'm fine with a usethis dependency if it solves a real problem that would be difficult to solve otherwise.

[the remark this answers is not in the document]
yutannihilation: I see.

R/use_extendr.R
Ilia-Kosenkov: You are correct, and it is still there. If you would like to keep the original behavior, writing only if `src/` is not present, then we can switch back to {brio}.

R/use_extendr.R
Ilia-Kosenkov: @clauswilke, @yutannihilation,
I think I am not entirely following the discission.
1. `rextendr::use_extendr()` creates wrappers (currently it writes `@useDynLib`)
2. `devtools::document()` compiles code, updates `NAMESPACE`, loads package.
3. But, to get **correct** `NAMESPACE`, we need to compile, then load package and `.Call` into wrapper generation function, which at this moment does not work -- `NAMESPACE` has not been updated and `useDynLib()` is missing. The alternative is not using `pkgload::load_all` but something like `pkgbuild::build()` and then manually loading the library with `dyn.load`, creating up to date wrappers, unloading, and then calling `devtools::document`, which will consume correct wrappers with all the necessary {roxygen2} tags. That way we do not need to write `NAMESPACE`.

If you have a better solution or at least a tiny reprex of how `rextendr:document()` should work, please share or make a PR in this branch.

yutannihilation: Can we dyn.load() the lib file directly?

Ilia-Kosenkov:
@clauswilke, {fs} is no longer required and does not really solve any problems or make life easier. We use both `fs::*` and `file.*` apis throughout the project, so replacing {fs} will make the codebase consistent.

R/use_extendr.R
Ilia-Kosenkov: @yutannihilation, yes. We basically do this in {rextendr} in `rust_source()` and similar functions, which generate libraries on the fly from user input ({Rcpp}/{cpp1} should do the same). The challenge is to reliably load and **unload**, otherwise, the dll will get stuck (at least on Windows).
There is `devtools::unload()` that I think unloads everything from everywhere. Alternatively, the fact we use {callr} means that by the end of the `callr::r()` another process has already finished, and the library is naturally unloaded.

yutannihilation: @Ilia-Kosenkov 
Here's my take:

https://github.com/Ilia-Kosenkov/rextendr/pull/2

I believe this is the same idea as what you suggested here (sorry, I typed the comment above before viewing this).

> 3\. The alternative is not using `pkgload::load_all` but something like `pkgbuild::build()` and then manually loading the library with `dyn.load`, creating up to date wrappers, unloading, and then calling `devtools::document`, which will consume correct wrappers with all the necessary {roxygen2} tags. That way we do not need to write `NAMESPACE`.

Ilia-Kosenkov: @yutannihilation,
Currently `pkgbuild::compile_dll` does not re-compile if `pkgbuild::needs_compile()` returns false.
I suspect there are other path related issues as well, which we do not uncover with the CI tests.
I am investigating this right now.

yutannihilation: Ah, I didn't notice it. Do you mean I should have added `force = TRUE`?

Ilia-Kosenkov: @yutannihilation, Yeah, check the last two commits here. If you have time, feel free to checkout this PR and try out new package creation & building.

R/use_extendr.R
clauswilke: I think that's a save approach for now. Check if `src` exists already, and if it does, don't do anything. It means we can make much stronger assumptions throughout the rest of the codebase.

R/ui.R
```diff
@@ -0,0 +1,90 @@
...
+        message,
+        details,
+        .sep = "\n"
+      ),
+      call. = FALSE
+    )
+  } else {
+    stop(message, call. = FALSE)
+  }
+}
```
clauswilke, line 90: There's a newline missing here.

clauswilke:
I'm starting to find it difficult to keep track of the various conversation threads. Are there any major showstoppers in this PR as it stands right now? (If something currently needs to be done manually that we'd want to automate, that's not a showstopper in my opinion. A showstopper to me would be a case where something is completely broken, deletes all your files, or prevents the package check from succeeding.) If not, I'd suggest we merge and then open separate issues for specific things still to resolve.

clauswilke:
I think relying on `dyn.unload()` is risky:

> The function dyn.unload unlinks the DLL. Note that unloading a DLL and then re-loading a DLL of the same name may or may not work: on Solaris it uses the first version loaded. Note also that some DLLs cannot be safely unloaded at all: unloading a DLL which implements C finalizers but does not unregister them on unload causes R to crash.

Ilia-Kosenkov:
@clauswilke,
Everything is in place except for the `dyn.unload.` We run `dyn.load` out of process. Do you prefer to rely on the fact the library is unloaded when the process ends? If yes, I can remove `dyn.unload` and we are ready to proceed.

R/ui.R
Ilia-Kosenkov: Done

clauswilke:
If it's out of process I have no concerns. Let's leave it in, as it's cleaner to unload explicitly.

clauswilke (approved):

PR #76, opened 2021-03-14
Removing {fs} dependency (Fixes #69)

### LINKED ISSUE #69: Remove fs dependency
My understanding is there is no major reason to keep fs so let's remove it to keep dependencies low.

Ilia-Kosenkov:
Removes {fs} in favor of base R (with some minor limitations). Fixes #69.

### DIFF
--- a/R/track_rust_source.R
+++ b/R/track_rust_source.R
@@ -2,15 +2,58 @@
 # E.g., outer_root/some_folder/code/packages/my_package/src/rust/src/lib.rs
 # becomes src/rust/src/lib.rs.
 # Used for pretty printing.
+# Assumes that `path` is within `package_root`.
+# @param path Scalar path to format.
+# @param search_from Path from which package root is looked up.
+# @returns `path`, relative to the package root.
 pretty_rel_path <- function(path, search_from = ".") {
-  fs::path_rel(path, start = rprojroot::find_package_root_file(path = search_from))
+  # Absolute path to the package root
+  package_root <- normalizePath(
+    rprojroot::find_package_root_file(path = search_from),
+    winslash = "/"
+  )
+
+  # Absolute path.
+  # `path` may not exists, so `mustWork` suppresses unnecessary warnings.
+  path <- normalizePath(path, winslash = "/", mustWork = FALSE)
+
+  # If `package_root` is not parent of `path`,
+  # return `path` unchanged (for simplicity)
+  if (!stringi::stri_detect_fixed(
+    str = path,
+    pattern = package_root,
+    case_insensitive = TRUE
+  )) {
+    return(path)
+  }
+
+  # If `path` is a subpath of `package_root`,
+  # then `path` contains `package_root` as a substring.
+  # This removes `pacakge_root` substring from `path`,
+  # performing comparison case_insensitively.
+  path <- stringi::stri_replace_first_fixed(
+    str = path,
+    pattern = package_root,
+    replacement = "",
+    case_insensitive = TRUE
+  )
+
+  # At this point, `path` can potentailly have a leading `/`
+  # Removes leading `/` if present.
+  path <- stringi::stri_replace_first_regex(path, "^/", "")
+
+  if (!nzchar(path)) {
+    path <- "."
+  }
+
+  path
 }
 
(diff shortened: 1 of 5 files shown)


R/track_rust_source.R
```diff
@@ -2,15 +2,58 @@
...
 pretty_rel_path <- function(path, search_from = ".") {
-  fs::path_rel(path, start = rprojroot::find_package_root_file(path = search_from))
+  # Absolute path to the package root
+  package_root <- normalizePath(
+    rprojroot::find_package_root_file(path = search_from),
+    winslash = "/"
+  )
+
+  # Absolute path.
+  # `path` may not exists, so `mustWork` suppresses unnecessary warnings.
```
clauswilke, line 17: "may not *exist*"

R/track_rust_source.R
```diff
@@ -111,3 +155,17 @@ needs_compilation <- function(path = ".", quiet = FALSE) {
   TRUE
 }
 
+# Replaces `fs::file_info`.
```
clauswilke, line 158: Instead of "Replaces", write "Equivalent to"?

clauswilke (approved):
Thanks!

clauswilke:
Two more comments:

1. I assume you've seen #79. It would be best to address this here.
2. This brings up the issue that for any non-trivial helper functions that we write, we should add some unit test functions that verify the function behavior. Here, for `pretty_rel_path()`, having a unit test that tests the relative path in a package and a unit test that tests the behavior outside of a package would be great. Similarly, one or two unit tests for `get_file_info()` would be useful too.

Ilia-Kosenkov:
@clauswilke,
I am not sure how to test `get_file_info` -- it basically calls `file.info` and wraps the output. 
To be able to *really* test it, we need to create a file with specific `ctime`/`mtime`/`atime`, which then can be read and its properties compared to the expected values. I do not see how we can achieve this with R without external dependencies.

The only base tool I found is [`Sys.setFileTime`](https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/Sys.setFileTime), which modified `mtime`. Perhaps it is enough for us, as `mtime` is the only field we really care about.

clauswilke:
I think you're overthinking this. `get_file_info()` is a wrapper and convenience function, so you test that functionality. Creat a temp, run both `get_file_info()` and `file.info()` on it, and test that the result is the same. Also test that the output has the format you expect, and that `get_file_info()` behaves reasonably on bad input, e.g. on a file that doesn't exist.

R/track_rust_source.R
```diff
@@ -1,16 +1,89 @@
...
+        pattern = package_root,
+        case_insensitive = TRUE
+      )
+  ) {
+    return(path)
+  }
+
+  # If `path` is a subpath of `package_root`,
+  # then `path` contains `package_root` as a substring.
+  # This removes `pacakge_root` substring from `path`,
```
clauswilke, line 62: spelling: `package_root`

clauswilke (approved):
Thanks, looks good. I found one typo, which I pointed out in a separate comment.

tests/testthat/test-helper-functions.R
```diff
@@ -94,3 +94,53 @@ test_that("`pretty_rel_path()` works", {
     "A/B/C/D.F"
   )
 })
+
+
+# Verifies that `rextendr::get_file_info()` returns the same as base R
+# `file.info()` for files and directories.
+test_that("`get_file_info()` and `file.info()` outptus are equivalent", {
```
clauswilke, line 101: typo: "output"

tests/testthat/test-helper-functions.R
```diff
@@ -94,3 +94,53 @@ test_that("`pretty_rel_path()` works", {
...
+  n_test_files <- 10L
+  n_test_dirs <- 2L
+  # Creates 10 temp files in 10 *different* temp directories,
```
clauswilke, line 104: Do we need 10? Maybe 3 is sufficient?

Ilia-Kosenkov:
Unless I made another typo somewhere, I think everything has been implemented & tested.

clauswilke:
Yes, please go ahead and merge. Thanks!

PR #81, opened 2021-03-15
Skip compiling and generating wrappers when unnecessary

### LINKED ISSUE #73: Document parameters `compile` in internal function `make_wrappers_externally()`
Currently this parameter is not documented:
https://github.com/extendr/rextendr/blob/74817048d578c640e77df740e1fa00071320a011/R/register_extendr.R#L120-L128

yutannihilation:
Close #73

Currently, the need for compile is determined in `rextendr::document()` and always `TRUE` or `FALSE` is passed to `register_extendr()` and `make_wrappers_externally()` accordingly. However, there is some need to improve this:

* `register_extendr()` should accept `compile = NA` to automatically determine the need for compilation
* `register_extendr()` should skip generating wrappers when it's not needed (c.f. suggested in https://github.com/extendr/rextendr/issues/73#issuecomment-799247721)

Therefore, this pull request proposes the following changes to achieve the above points:

* Move the logic determining whether to compile to `register_extendr()`.
* Move the compilation from `make_wrappers_externally()` to `register_extendr()` (This is consistent with the fact that `make_wrappers()` doesn't compile).
* Handle `compile = NA` properly in `register_extendr()`.
* Skip generating wrappers when the wrapper file is newer than the DLL; of course, this isn't always the case (e.g., the user might mistakenly modify the wrapper file), so provide `force` arg to force regeneration.
* Remove `force_wrappers` arg because this is no longer needed after #67 made `make_wrappers_externally` independent from `NAMESPACE`, and it's might be confused with `force` arg.


yutannihilation:
@Ilia-Kosenkov 
I need to add more tests, but I believe I basically implemented what you suggested in https://github.com/extendr/rextendr/issues/73#issuecomment-799247721:

> I would go even further and introduce a flag (and re-introduce the function we deprecated earlier) that determines if wrappers should be re-generated. Currently, every call to rextendr::document() overwrite wrppers, which is not really needed.

But I'm still not sure what "the function we deprecated earlier" is so I might misunderstand the context... Could you take a look when you have time and give some feedback?

R/register_extendr.R
```diff
@@ -40,27 +45,38 @@ register_extendr <- function(path = ".", quiet = FALSE, force_wrappers = FALSE,
...
+  # by hand) so the user might need to run with `force = TRUE`.
+  if (!isTRUE(force) &&
+      isTRUE(file.exists(dll_file)) &&
+      isTRUE(file.exists(outfile)) &&
+      isTRUE(file.info(dll_file)$mtime < file.info(outfile)$mtime)) {
+    cli::cli_alert_info("{.file {outfile}} is up-to-date. Skip generating wrapper functions.")
+    cli::cli_alert_info("If you want to force the generation, run `register_extendr(force = TRUE)`.")
+
+    return(invisible(character(0L)))
   }
```
Ilia-Kosenkov, line 80: I was talking about `needs_new_wrappers` function that essential compared `mtime`s of the dll and wrapper files.
Now this looks like a little bit of repetition to me.

Would you consider combining this snippet with `needs_compilation()` method, extracting file `mtime` comparison logic into another helper?
E.g.
```R
was_modified_ater <- function(file_paths, compare_to) {
}
```
Here `file_paths` is a paths (character) vector of any length, `compare_to` is a scalar path, to which comparison happens. The output is a logical vector of `length("file_paths")`, returning `TRUE` when `paths$mtime > compare_to$mtime` (pseudo code), `FALSE` otherwise, `NA` if path is invalid or input was `NA`.

This will allow to write something like:
```R
needs_compilation <- function(path, quiet) }
  was_modified_after(get_rust_files(path), get_library_path(path))
  # Verbose message goes here
}
```

And replace this snippet with something like:

```R
was_modified_after(dll_path, out_file)
```

R/register_extendr.R
```diff
@@ -40,27 +45,38 @@ register_extendr <- function(path = ".", quiet = FALSE, force_wrappers = FALSE,
...
+      path = path,
+      force = TRUE,
+      quiet = quiet
+    )
+  }
+
+  dll_file <- get_library_path(path)
+
+  # If the wrapper file is newer than the DLL file, assume it's generated by
+  # the latest DLL, which should mean it doesn't need regenerated. This isn't
```
Ilia-Kosenkov, line 69: *doesn't need to **be** regenerated*

R/register_extendr.R
```diff
@@ -40,27 +45,38 @@ register_extendr <- function(path = ".", quiet = FALSE, force_wrappers = FALSE,
...
+  # always the case (e.g. when the user accidentally edited the wrapper file
+  # by hand) so the user might need to run with `force = TRUE`.
+  if (!isTRUE(force) &&
+      isTRUE(file.exists(dll_file)) &&
+      isTRUE(file.exists(outfile)) &&
+      isTRUE(file.info(dll_file)$mtime < file.info(outfile)$mtime)) {
+    cli::cli_alert_info("{.file {outfile}} is up-to-date. Skip generating wrapper functions.")
```
Ilia-Kosenkov, line 76: I think it is better to have `{.file {pretty_rel_path(outfile)}}` than just `outfile`. {rprojroot} should produce an absolute path, so `outfile` is very long. `{pretty_rel_path(outfile)}` outputs path relative to the package root, so something like `R/extendr-wrappers.R`.

R/register_extendr.R
```diff
@@ -40,27 +45,38 @@ register_extendr <- function(path = ".", quiet = FALSE, force_wrappers = FALSE,
...
+    cli::cli_alert_info("If you want to force the generation, run `register_extendr(force = TRUE)`.")
```
Ilia-Kosenkov, line 77: 1. I suggest using `cli::cli_alert_warning` to draw the user's attention, as this message is important.
2. I suggest `"..., run {.code rextendr::register_extendr(force = TRUE)},"` for consistency, see [{cli} inline markup](https://cli.r-lib.org/reference/inline-markup.html). Note that `{.fun ...}` will not work, as it appends an extra parentheses to the function name.

Ilia-Kosenkov: On second thought, do we really need to write this every time `rextendr::document()` is called without any changes to the Rust code? I'd say this is a standard situation.

yutannihilation: Thanks, I'm not yet familiar with cli's expressions. Let's use it.

> On second thought, do we really need to write this every time `rextendr::document()` is called without any changes to the Rust code?

Do we want to warn once a session? Or, do we want to stop warning at all?

R/register_extendr.R
yutannihilation: Agreed, I'll implement it.

Ilia-Kosenkov: I am not sure about naming. If you have a better name than `was_modified_after`, please use it.

R/register_extendr.R
Ilia-Kosenkov: It feels like we do not need it all. @clauswilke, what do you think?

Ilia-Kosenkov:
I think we need to pass `path` to `pretty_rel_path`: `pretty_rel_path(outfile, path)`, it will not work correctly if `register_extendr` is invoked from outside of the package directory.

It also seems like some issue with distribution R for Ubuntu, which causes all Linux CIs to fail.

R/register_extendr.R
clauswilke: I always get annoyed when code tells me stuff I know already. Even if only once a session. I say remove it. People can read the documentation or ask a question on stackoverflow if they truly don't know how to force the generation of the wrappers.

R/register_extendr.R
```diff
@@ -10,16 +10,21 @@
...
-#' @inheritParams pkgload::load_all
 #' @param path Path from which package root is looked up.
 #' @param quiet Logical indicating whether any progress messages should be
 #'   generated or not.
-#' @param force_wrappers Logical indicating whether to install a minimal wrapper
-#'   file in the cases when generating wrappers by Rust code failed. This might
-#'   be needed when the wrapper file is accidentally lost or corrupted.
+#' @param force Logical indicating whether to force re-generating
+#'   `R/extendr-wrappers.R` even when it doesn't seem to need updated (By
+#'   default, generation is skipped when it's newer than the DLL).
```
clauswilke, line 18: Just a minor quibble: either use lower case "by" or put a period before the parentheses and a second one inside: "...need to be updated. (By default, ... the DLL.)"

R/register_extendr.R
```diff
@@ -40,27 +45,37 @@ register_extendr <- function(path = ".", quiet = FALSE, force_wrappers = FALSE,
...
+    pkgbuild::compile_dll(
+      path = path,
+      force = TRUE,
+      quiet = quiet
+    )
+  }
+
+  dll_file <- get_library_path(path)
+
+  # If the wrapper file is newer than the DLL file, assume it's generated by the
```
clauswilke, line 68: Write "assume it's been generated" to make the chain of causality absolutely clear?

R/register_extendr.R
```diff
@@ -72,10 +87,16 @@ register_extendr <- function(path = ".", quiet = FALSE, force_wrappers = FALSE,
...
       use_symbols = TRUE,
-      quiet = quiet,
-      compile = compile
+      quiet = quiet
     ),
-    error = error_handle
+    error = function(e) {
+      ui_throw(
+        "Failed to generate wrapper functions.",
+        c(
```
clauswilke, line 95: What is the purpose of the `c()` here? Is it needed? If yes, maybe add a comment explaining what it does?

Ilia-Kosenkov: This (likely) was done by me after introducing `ui_*` functions, and I forgot to remove it when the number of additional displayed messages changed.  `c()` can be safely removed in this case.

R/register_extendr.R
yutannihilation: I chose `find_newer_files_than()`, as it needs to return character vector to show the message here:

https://github.com/extendr/rextendr/blob/def4813dd003efaa9ea7b7b03cbc625c22dcd651/R/track_rust_source.R#L178-L183

yutannihilation:
Thanks for your reviews! I believe now I addressed all the comments and added minimal tests.

clauswilke:
Could you resolve the git conflicts?

tests/testthat/test-helper-functions.R
```diff
@@ -144,3 +144,26 @@ test_that("`get_file_info()` and `file.info()` outputs are equivalent", {
   expect_true(is.na(na_info[["ctime"]]))
   expect_true(is.na(na_info[["atime"]]))
 })
+
+test_that("find_newer_files_than() works", {
+  old_file <- tempfile()
+  new_file1 <- tempfile()
+  new_file2 <- tempfile()
+  brio::write_file("", old_file)
+  Sys.sleep(1)
```
clauswilke, line 153: One second is a long time. Does .01 work? What's the lowest resolution for mtime?

R/register_extendr.R
```diff
@@ -40,29 +45,51 @@ register_extendr <- function(path = ".", quiet = FALSE, force_wrappers = FALSE,
...
+      ui_throw(msg)
+    } else {
+      # If compile wasn't invoked, it might succeed with explicit "compile = TRUE"
       ui_throw(
-        "Failed to generate wrapper functions.",
-        c(
-          ui_x(e[["message"]])
-        )
+        msg,
+        ui_i("You need to compile first, try {.code `register_rextendr(compile = TRUE)`}")
```
Ilia-Kosenkov, line 77: Unfortunately, this will not work.
For some reason {cli} cannot be used directly to format strings, it always emits `message`s.
I believe this is by design but I would very much like to get an {cli}-interpolated `ansi_string`.
To avoid problems, remove `{.code ...}` and just write ``` `register_extendr(...)` ```.

If I find a solution, I will improve implementation of `ui_*` functions.

yutannihilation: Oh, thanks. I'm yet to figure out how to use these functions...

tests/testthat/test-helper-functions.R
yutannihilation: > What's the lowest resolution for mtime?

I couldn't find the answer, but it seems `.01` works.

yutannihilation:
Hmm, Bintray problem again...

R/track_rust_source.R
```diff
@@ -200,3 +184,32 @@ get_file_info <- function(path) {
...
+# @param reference A file path to compare `files` against.
+# @return File paths newer than `reference`.
+find_newer_files_than <- function(files, reference) {
+  # If `files` is of length 0 or NULL, exit early.
+  if (length(files) == 0L) {
+    return(character(0))
+  }
+
+  if (length(reference) != 1L) {
+    stop("`reference` must be length 1 of character vector", call. = FALSE)
```
Ilia-Kosenkov, line 200: Same with `stop` here. We could provide super-detailed error messages now.
```R
details <- character(0)
if (length(reference) != 1L) {
  details <- ui_x("Expected vector of length `1`, got `{length(reference)}`.")
}
if(typeof(reference) != "character") {
  details <- c(details, ui_x("Expected type `character`, got `{typeof(reference)}`."))
}
ui_throw("Invalid argument `reference`.", details = details)
```

R/track_rust_source.R
```diff
@@ -200,3 +184,32 @@ get_file_info <- function(path) {
...
+  }
+
+  if (is.na(reference) || !file.exists(reference)) {
+    stop("File ", reference, " doesn't exist", call. = FALSE)
```
Ilia-Kosenkov, line 204: And possibly here.

R/track_rust_source.R
yutannihilation: Thanks, I was wondering how detailed the error here should be, as this is more like an assertion that should be always met internally, not an user-facing error. No matter however friendly the error message is, it's probably not what user can fix.

clauswilke (approved):
Looks good to me. No further comments from my side.

yutannihilation:
@Ilia-Kosenkov 
Do you have any additional comments? If this looks good to you too, I'm going to merge this.

Ilia-Kosenkov (approved):
Looks good, please merger as soon as you are ready.

yutannihilation:
Thanks!

PR #82, opened 2021-03-15
Add @noRd to non-exported functions

### LINKED ISSUE #71: Don't generate .Rd pages for internal functions
A few internal functions are currently documented using roxygen2, so man pages are generated. I think we should turn off man page generation. This generally means replacing `#'` with `#` at the beginning of each line.

These are the cases I'm aware of:

https://github.com/extendr/rextendr/blob/74817048d578c640e77df740e1fa00071320a011/R/register_extendr.R#L85-L100

https://github.com/extendr/rextendr/blob/74817048d578c640e77df740e1fa00071320a011/R/register_extendr.R#L120-L126

yutannihilation:
Close #71


clauswilke (approved):
Thanks!

PR #92, opened 2021-03-19
Move devtools to Suggest

### LINKED ISSUE #90: Remove {devtools} dependency
We should remove {devtools} from imports. It's a pretty heavy dependency, and we use it only to call {roxygen2}. We could do that directly.

See also: https://github.com/extendr/rextendr/issues/88#issuecomment-802625576

yutannihilation:
Fix #90 

While I understand we want to get rid of the heaviness of devtools' dependency chain, I think it's fair to expect that there's devtools installed on the machine of such users who wants to use `rextendr::document()` functionality. The advantage of using `devtools::document()` directly over implementing it by ourselves is that we don't need to track the changes of the behavior of `devtools::document()`. While it's not a very huge function, I'm afraid we would suffer from the subtle differences between the real `devtools::document()` and our version otherwise.

Special thanks to @malcolmbarrett for letting me know about `check_installed()` :)


clauswilke (approved):
Looks good. (Assuming checks will pass.)

yutannihilation:
Thanks.

PR #93, opened 2021-03-22
Remove {usethis}

### LINKED ISSUE #70: Remove usethis dependency
We currently only use usethis for `usethis::write_over()`. I think we can remove this dependency, in particular if we take the position that we don't write any files if `src` already exists.

Ilia-Kosenkov:
Fixes #70. Removes explicit usage of {usethis} (mainly, `usethis::write_over()`).
Introduces `rextendr::write_file()`, a wrapper around `brio::write_lines()` that displays verbose messages if operation succeeds.

### DIFF
--- a/R/write_file.R
+++ b/R/write_file.R
@@ -1,7 +1,7 @@
 #' Writes text to file
 #'
-#' This function is a wrapper around [`brio::write_lines`].
-#' It also supports verboe output, similar to [`usethis::write_over`],
+#' This function is a wrapper around [`brio::write_lines()`].
+#' It also supports verboe output, similar to [`usethis::write_over()`],
 #' controlled by the `quiet` parameter.
 #' @param text Character vector containing text to write.
 #' @param path A string giving the file path to write to.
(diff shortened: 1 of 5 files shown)


R/write_file.R
```diff
@@ -0,0 +1,22 @@
+#' Writes text to file
+#'
+#' This function is a wrapper around [`brio::write_lines()`].
+#' It also supports verboe output, similar to [`usethis::write_over()`],
```
clauswilke, line 4: typo: "verbose"

R/write_file.R
```diff
@@ -0,0 +1,22 @@
...
+#' controlled by the `quiet` parameter.
+#' @param text Character vector containing text to write.
+#' @param path A string giving the file path to write to.
+#' @param search_root_from A string giving path to a package subfolder,
+#' which is used to find package root and produce message for the user.
```
clauswilke, line 9: I don't entirely understand what this parameter does from this documentation. Is this only for message generation? If yes, say so explicitly. "This parameter only affects any messages generated to the user. It has no effect on where the file is written."

clauswilke (approved):
Looks good to me. I have two minor comments on the documentation that should be fixed before merging.

Ilia-Kosenkov:
Done. Please check if it reads better now.

clauswilke:
Yes, looks good. Thanks!

PR #94, opened 2021-03-31
Update ui functions

malcolmbarrett:
This is the first of two PRs that I plan to submit to address #62. Doesn't quite close that issue.

This PR:
* Updates all user communication to use `ui_*()` functions
* Refactors `ui_*()` functions to pipe through `rlang::inform()` and to use the new internal function `bullet()` to format text via cli.
* Refactors `ui_throw()` to simplify its logic and use `rlang::abort()` to throw an error of class `rextendr_error`
* Renames and modifies `ui_q()` to `ui_o()` to be more similar to `usethis::ui_todo()`, which is usually how questions are asked in usethis
* Allows global UI silencing via `usethis.quiet = FALSE`. I did this to be consistent with usethis, although you may prefer a custom option like `rextendr.quiet`.
* Updates tests. Generally, these are simplified versions of previous message tests since unicode is a nightmare to handle in regex. I think they are good enough tests for their purposes. I also introduced `expect_rextendr_error()` to more strictly expect an error thrown by rextendr.
* I also ran `use_tidy_description()` 🧹 😄 

Shoutout to @Ilia-Kosenkov who laid the groundwork for this PR!

### DIFF
--- a/R/register_extendr.R
+++ b/R/register_extendr.R
@@ -37,8 +37,8 @@ register_extendr <- function(path = ".", quiet = FALSE, force = FALSE, compile =
     ui_throw(
       "Unable to register the extendr module.",
       c(
-        bullet("x", "Could not find file {.file src/entrypoint.c }."),
-        bullet("o", "Are you sure this package is using extendr Rust code?")
+        bullet_x("Could not find file {.file src/entrypoint.c }."),
+        bullet_o("Are you sure this package is using extendr Rust code?")
       )
     )
   }
@@ -74,7 +74,7 @@ register_extendr <- function(path = ".", quiet = FALSE, force = FALSE, compile =
       # If compile wasn't invoked, it might succeed with explicit "compile = TRUE"
       ui_throw(
         msg,
-        bullet("i", "You need to compile first, try {.code register_rextendr(compile = TRUE)}.")
+        bullet_i("You need to compile first, try {.code register_rextendr(compile = TRUE)}.")
       )
     }
   }
@@ -104,7 +104,7 @@ register_extendr <- function(path = ".", quiet = FALSE, force = FALSE, compile =
     error = function(e) {
       ui_throw(
         "Failed to generate wrapper functions.",
-        bullet("x", e[["message"]])
+        bullet_x(e[["message"]])
       )
     }
   )
--- a/R/register_extendr.R
+++ b/R/register_extendr.R
@@ -38,7 +38,7 @@ register_extendr <- function(path = ".", quiet = FALSE, force = FALSE, compile =
       "Unable to register the extendr module.",
       c(
         ui_x("Could not find file {.file src/entrypoint.c }."),
-        ui_q("Are you sure this package is using extendr Rust code?")
+        ui_o("Are you sure this package is using extendr Rust code?")
       )
     )
   }
--- a/R/register_extendr.R
+++ b/R/register_extendr.R
@@ -29,7 +29,7 @@ register_extendr <- function(path = ".", quiet = FALSE, force = FALSE, compile =
   pkg_name <- x$get("Package")
 
   if (!isTRUE(quiet)) {
-    cli::cli_alert_info("Generating extendr wrapper functions for package: {.pkg {pkg_name}}.")
+    ui_i("Generating extendr wrapper functions for package: {.pkg {pkg_name}}.")
   }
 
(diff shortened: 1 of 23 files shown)


R/register_extendr.R
```diff
@@ -38,7 +38,7 @@ register_extendr <- function(path = ".", quiet = FALSE, force = FALSE, compile =
       "Unable to register the extendr module.",
       c(
         ui_x("Could not find file {.file src/entrypoint.c }."),
-        ui_q("Are you sure this package is using extendr Rust code?")
+        ui_o("Are you sure this package is using extendr Rust code?")
```
Ilia-Kosenkov, line 41: Should this be some version of `bullet()` instead of `ui_*`?

clauswilke:
Could you add a file with developer guidelines summarizing how these functions should be used? Along the lines of this file in extendr: https://github.com/extendr/extendr/blob/master/MAINTAINERS_GUIDE.md

R/ui.R
```diff
@@ -1,11 +1,33 @@
...
+}
+
+ui_bullet <- function(type, text, env = parent.frame()) {
+  if (getOption("usethis.quiet", FALSE)) {
+    return(invisible())
+  }
+
+  rlang::inform(bullet(type, text, env = env))
+}
+
```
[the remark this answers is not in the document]
malcolmbarrett: I considered that approach, too, so definitely fine with it!

R/register_extendr.R
malcolmbarrett: yes, missed these -- thanks!

malcolmbarrett:
@clauswilke what do you think about putting it in a `CONTRIBUTING.md` generated by `usethis::use_tidy_contributing()`?

clauswilke:
We have contributing guidelines over at extendr and they could be adapted:
https://github.com/extendr/extendr/blob/master/CONTRIBUTING.md

However, I think that's not the place for documentation that would cover internal coding conventions. I'm not sure where that would go, though. Most projects don't have such documentation and it's often a major problem, as nobody knows how exactly code should be written.

malcolmbarrett:
Yeah, agreed. usethis has a helpful document along those lines here: https://github.com/r-lib/usethis/blob/master/principles.md and in fact has a [section on communicating with the user](https://github.com/r-lib/usethis/blob/master/principles.md#communicating-with-the-user)

clauswilke:
Yes, exactly, something like this.

malcolmbarrett:
Ok, I added a `principles.md` document that includes details on UI and errors. I'll file an issue to include other items outside the scope of this PR

clauswilke (commented):
The `principles.md` file looks good to me. I haven't looked over all the other changes as I'm super busy right now.

@Ilia-Kosenkov Can you finish this off?

principles.md
```diff
@@ -0,0 +1,37 @@
...
+
+This guide documents important internal coding conventions used in rextendr.
+
+## Communicating with the user
+
+rextendr uses the cli package to format messages to the user, which are generally then routed through `rlang::inform()`. Most UI will happen through `ui_*()` functions:
+
+| function | purpose                                                                                                       |
+|----------|---------------------------------------------------------------------------------------------------------------|
+| `ui_v()` | communicate that rextendr has done something, such as write a file                                            |
```
Ilia-Kosenkov, line 11: I suggest adding "... has done something *successfully*, ..." to emphasize that `*_v` functions communicate positive results.

malcolmbarrett: yes, good point -- updated!

Ilia-Kosenkov (approved):
Looks good to me. Thanks!

clauswilke:
One more thing that I can't check easily from my phone right now: did you add the principles file to .Rbuildignore?

malcolmbarrett:
yes, already done

Ilia-Kosenkov:
Is there anything blocking this PR?

clauswilke:
Not from my end.

Ilia-Kosenkov:
@malcolmbarrett , you should probably merge this.

malcolmbarrett:
I can't. I don't have access 😎

Ilia-Kosenkov:
@malcolmbarrett, oh shoot, I had no idea, I am sorry. If you think it is ready, I can merge it.

malcolmbarrett:
No worries! Yes, I think it is ready

malcolmbarrett:
Thanks!

PR #97, opened 2021-03-31
Refactor pkg name calls

malcolmbarrett:
This little PR refactors a bit of internal code that calls desc in several places in an identical manner to instead use a helper function `pkg_name()`


clauswilke (approved):
Thanks!

Issue #95, opened 2021-03-31
Flesh out `principles.md`


malcolmbarrett:
#94 adds `principles.md` to make explicit important internal coding conventions used in rextendr. That PR limited the contents of `principles.md` to user communication and errors. However, there are probably other conventions (or emerging conventions waiting to be codified) that would do well to be included.

Two particularly good examples of documents like this are in [usethis](https://github.com/r-lib/usethis/blob/master/principles.md) and [targets](https://books.ropensci.org/targets-design/index.html). These are both very well designed packages, and I think the principled approach to design helped get them there.

Ilia-Kosenkov:
I suggest writing about `to_toml()` and its use cases.
I strongly encourage using it every time we write a `*.toml` file. I believe it supports the majority of use cases that arise in Rust projects (it does not fully support TOML standard though).

malcolmbarrett:
It seems like it might be good to document how to handle paths, e.g. when to use `pretty_rel_path()`. That's something usethis is quite explicit about in its `principles.md`.

clauswilke:
Also, per #102, describe snapshot testing.

PR #102, opened 2021-04-15
Update tests, bump to testthat 3

### LINKED ISSUE #98: Upgrade to testthat 3
testthat edition 3 requires that you opt-in. Since this project is young, I think it would be wise to do so now for the long-term health of the codebase. It's possible that will require some refactoring in the tests or code, but testthat 3 also comes with [some nice features](https://testthat.r-lib.org/articles/third-edition.html)

malcolmbarrett:
This PR refactors the tests in several ways:

1. Bump the testthat version to edition 3 and use parallel tests. This turned out to be easy, as none of the tests failed after 😎 
2. Use `usethis.quiet` introduced in #94 to quiet all tests prior to running (I bumped into noisy tests running them in certain ways) and reset options during teardown
3. Adds withr and usethis to Suggests, which were already implicitly suggested via devtools. 
4. Uses withr to manage local states, e.g. temp files or options
5. Uses withr and usethis to create local packages via `local_package()`. This is a [self-cleaning test fixture](https://www.tidyverse.org/blog/2020/04/self-cleaning-test-fixtures/).
6. Adds direct tests for `use_extendr()`

I thought about more generally improving test coverage, but this seems reasonable for now and will let me finish #62 a bit more easily

Closes #98

### DIFF
--- a/tests/testthat/helper.R
+++ b/tests/testthat/helper.R
@@ -1,3 +1,36 @@
 expect_rextendr_error <- function(...) {
   expect_error(..., class = "rextendr_error")
 }
+
+local_temp_dir <- function(envir = parent.frame()) {
+  current_wd <- getwd()
+  path <- tempfile()
+  dir.create(path)
+
+  setwd(path)
+
+  withr::defer(
+    {
+      setwd(current_wd)
+      usethis::proj_set(NULL)
+      unlink(path)
+    },
+    envir = envir
+  )
+
+  invisible(path)
+}
+
+local_proj_set <- function(envir = parent.frame()) {
+  old_proj <- usethis::proj_set(getwd(), force = TRUE)
+  withr::defer(usethis::proj_set(old_proj), envir = envir)
+}
+
+local_package <- function(nm, envir = parent.frame()) {
+  local_temp_dir(envir = envir)
+  dir <- usethis::create_package(nm)
+  setwd(dir)
+  local_proj_set(envir = envir)
+
+  invisible(dir)
+}
--- a/tests/testthat/helper.R
+++ b/tests/testthat/helper.R
@@ -14,7 +14,7 @@ expect_rextendr_error <- function(...) {
 #' It also sets the local working directory and usethis project to the temporary
 #' package. These settings are reverted and the package removed via
 #' `withr::defer()`. This clean-up happens at the end of the local scope,
-#' usually the end of a `testthat()` call.
+#' usually the end of a `test_that()` call.
 #'
 #' @param nm The name of the temporary package
 #' @param envir An environment where `withr::defer()`'s exit handler is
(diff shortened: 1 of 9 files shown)


clauswilke:
Could you add some comments to your code? In particular, I think every function (including helper functions) should have a brief explanation of what it does and what its arguments are.

malcolmbarrett:
Sure, I added some roxygen comments to the helper functions

clauswilke (approved):
Thanks, looks good to me.

@Ilia-Kosenkov Could you take a look also please?

tests/testthat/helper.R
```diff
@@ -1,3 +1,75 @@
...
+#' scope.
+#'
+#' @param envir An environment where `withr::defer()`'s exit handler is
+#'   attached, usually the `parent.frame()` to exist locally
+#'
+#' @return A path to the temporary directory
+local_temp_dir <- function(envir = parent.frame()) {
+  current_wd <- getwd()
+  path <- tempfile()
+  dir.create(path)
```
Ilia-Kosenkov, line 47: Is there a reason `tempfile()` is used instead of `tempdir()`?

Ilia-Kosenkov:
Perhaps we should also add a note for developers about how to use snapshot testing in the future.

tests/testthat/helper.R
yutannihilation: `tempdir()` is a function that returns the parent directory of all temporary files, which contains various temporary files. So, if we want a clean temporary directory, `dir.create(tempfile())` is what we want.

``` r
tempdir()
#> [1] "/tmp/RtmpbzxQ5a"
tempfile()
#> [1] "/tmp/RtmpbzxQ5a/file185566cc1191"
```

<sup>Created on 2021-04-16 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

Ilia-Kosenkov: I see, thanks @yutannihilation !

Ilia-Kosenkov (approved):

tests/testthat/helper.R
malcolmbarrett: Yeah, those names are a bit misleading IMO. I have learned the hard way that `tempdir()` is actually a very important directory, and if you mess with it too much (or delete it 😇) you get all sorts of bizarre errors in R

malcolmbarrett:
@Ilia-Kosenkov Yeah, maybe a good thing to add to #95 (generally adding a testing section that mentions the functions implemented here + strategy)

clauswilke:
@malcolmbarrett Is this ready to merge?

malcolmbarrett:
Yes, unless there is anything else, it's ready on my end!

clauswilke:
Thanks!

PR #106, opened 2021-04-28
Use templates in `use_extendr()`

### LINKED ISSUE #62: Suggesting or replicating usethis functions

malcolmbarrett:
This PR refactors `use_extendr()` a bit more to use templates in the style of `usethis::use_template()`. When usethis is available, we just use that, and when it's not, we use a simplified version (`use_template()` is more robust but some of the protections that it has aren't strictly necessary here since we don't write at all if there are pre-existing files). 

Using this approach is more consistent with usethis, but I also find that templates are much easier to understand and maintain than hard-coded text in code.

This PR also adds mockr to `Suggests` to enable mocking of functions in tests. 

Finally closes #62

### DIFF
--- a/R/use_extendr.R
+++ b/R/use_extendr.R
@@ -195,6 +195,45 @@ extendr_module! {{
   return(invisible(TRUE))
 }
 
+use_rextendr_template <- function(template, save_as = template, data = list()) {
+  if (is_installed("usethis")) {
+    created <- usethis::use_template(
+      template,
+      save_as = save_as,
+      data = data,
+      open = FALSE,
+      package = "rextendr"
+    )
+
+    return(invisible(created))
+  }
+
+  template_path <- system.file(
+    "templates",
+    template,
+    package = "rextendr",
+    mustWork = TRUE
+  )
+
+  template_content <- brio::read_file(template_path)
+
+  template_content <- glue::glue_data(
+    template_content,
+    .envir = data,
+    .open = "{{{", .close = "}}}"
+  )
+
+  ui_v("Writing {.file save_as}")
+  brio::write_file(template_content, path = save_as)
+
+  invisible(TRUE)
+}
+
+#' Wrap `rlang::is_installed()` for ease of mocking installed packages
+is_installed <- function(pkg) {
+  rlang::is_installed(pkg)
+}
+
 #' Creates example R wrappers for Rust functions.
 #'
 #' Writes simple R wrappers, without which initial compilation of the
--- a/R/use_extendr.R
+++ b/R/use_extendr.R
@@ -34,7 +34,7 @@ use_extendr <- function(path = ".", quiet = getOption("usethis.quiet", FALSE)) {
     return(invisible(FALSE))
   }
 
-  rust_src_dir <- file.path("src", "rust", "src")
+  rust_src_dir <- file.path(src_dir, "rust", "src")
   dir.create(rust_src_dir, recursive = TRUE)
   ui_v("Creating {.file {pretty_rel_path(rust_src_dir, path)}}.")
 
--- a/R/use_extendr.R
+++ b/R/use_extendr.R
@@ -15,7 +15,6 @@
 #' @return A logical value (invisible) indicating whether any package files were
 #' generated or not.
 #' @export
-
 use_extendr <- function(path = ".", quiet = getOption("usethis.quiet", FALSE)) {
   pkg_name <- pkg_name(path)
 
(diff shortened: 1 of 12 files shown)


tests/testthat/_snaps/use_extendr.md
```diff
@@ -4,14 +4,125 @@
...
+      v Writing 'src/Makevars.win'
+      v Writing 'src/.gitignore'
+      v Writing 'src/rust/Cargo.toml'.
+      v Writing 'src/rust/src/lib.rs'
+      v Writing 'R/extendr-wrappers.R'
       v Finished configuring extendr for package testpkg.
       * Please update the system requirement in DESCRIPTION file.
       * Please run `rextendr::document()` for changes to take effect.
 
+---
```
malcolmbarrett, line 18: I also added snapshot tests (below) of the templates to make it easier to know that I wasn't breaking anything in this transition

clauswilke:
Let me know when this is ready to review and I'll take a look then. One more week of classes and then I'll be able to focus a bit more on rextendr again, and hopefully submit to CRAN.

malcolmbarrett:
Thanks! Ready to review. Had some weird issues with line endings on Windows, but it's resolved. 

That sounds great! Good luck with the end of the semester

R/use_extendr.R
```diff
@@ -195,43 +101,56 @@ extendr_module! {{
...
-      c(wrappers_content, extra_items),
-      sep = "\n"
+use_rextendr_template <- function(template, save_as = template, data = list(), quiet = getOption("usethis.quiet", FALSE)) {
+  if (is_installed("usethis")) {
+    created <- usethis::use_template(
+      template,
+      save_as = save_as,
+      data = data,
+      open = FALSE,
+      package = "rextendr"
```
Ilia-Kosenkov, line 120: How will `use_template` behave if we pass `quiet = TRUE` to `use_extendr_template` when the global option `usethis.quiet = FALSE`?
Is it possible that without {usethis} our implementation will be quiet, but **with** {usethis} it will be verbose?
If yes, perhaps we should wrap call to `use_template` in some `with*` method to explicitly set up {usethis} `quiet` option for the duration of its execution?

malcolmbarrett: Yes, you're right about that. Good catch! Luckily we can safely assume withr is installed since it's a dep of usethis, so that makes it easy.

R/use_extendr.R
```diff
@@ -195,43 +101,57 @@ extendr_module! {{
...
-    )"
-  )
-
-  if (!is.null(extra_items)) {
-    wrappers_content <- glue::glue_collapse(
-      c(wrappers_content, extra_items),
-      sep = "\n"
+use_rextendr_template <- function(template, save_as = template, data = list(), quiet = getOption("usethis.quiet", FALSE)) {
+  if (is_installed("usethis")) {
+    withr::local_options(usethis.quiet = quiet)
```
Ilia-Kosenkov, line 115: I noticed that {withr} is not in the `Imports`, but it is used in the source. 
I think we should add the reference. 
In the upcoming PR I am relying on {withr} as well, so it is here to stay.

clauswilke: I think the argument was that usethis uses withr and so if usethis is installed then withr is also installed. It's a bit shaky, though, since usethis could in theory stop using withr. I'm fine with moving it to `Imports`.

Ilia-Kosenkov: If I understand correctly, it should be OK because we depend on {pkgbuild} 
https://github.com/extendr/rextendr/blob/5ee2062d9134084a99e68266271985e360244d6e/DESCRIPTION#L42-L49
and it in turn depends on {withr}
https://github.com/r-lib/pkgbuild/blob/2aaecf838eb023b788b7a31e267ac644d616dae0/DESCRIPTION#L20.

malcolmbarrett: I'm not concerned about usethis switching away from withr (it's central to its internal state principles, and I'm also an author over there now, so I'd have a pretty good heads up), but let's just move it to Imports since we implicitly do so anyway. We can probably make use of it in a few other places, as well.

clauswilke:
@Ilia-Kosenkov If you have reviewed this carefully and are happy with it then please approve. I don't want to hold this up.

Ilia-Kosenkov:
@clauswilke, it seems Ok to me.
However, I would like to ask @malcolmbarrett to update our guidelines with info on snapshot testing and templates.
Perhaps, as a follow-up PR.

Ilia-Kosenkov (approved):

Ilia-Kosenkov:
@malcolmbarrett, should I merge this?

PR #112, opened 2021-05-10
Fixing incorrect arguments passed to invoke_cargo() and system2()

### LINKED ISSUE #111: Incorrect argument passed to `system2` call in `invoke_cargo`
I was looking for a way to capture `cargo` output when the compilation errors when I found this:
https://github.com/extendr/rextendr/blob/942d0d3b749ee7c957a003631aa7e9658c3be24f/R/source.R#L246-L259
`stdout` passed to both `stdout` and `stderr`.
The function, however, has both `stdout` and `stderr` arguments
https://github.com/extendr/rextendr/blob/942d0d3b749ee7c957a003631aa7e9658c3be24f/R/source.R#L223-L224

The wild thing is the caller of `invoke_cargo` passes to `stderr` an invalid value of `stderr`, which is **not assigned** in the caller method.
https://github.com/extendr/rextendr/blob/942d0d3b749ee7c957a003631aa7e9658c3be24f/R/source.R#L164-L172

Basically, we have two bugs that cancel each other because of the lazy evaluation of function arguments in R...

Ilia-Kosenkov:
Fixed #111.
Fixes a bug introduced by copy-pasting/moving code around in #38 by me.

The name of the variable that controls the output of `system2` is changed from confusing `stdout` to simply `out`.
It is passed to both `stdout` and `stderr`.

### DIFF
--- a/R/source.R
+++ b/R/source.R
@@ -219,7 +219,23 @@ rust_function <- function(code, env = parent.frame(), ...) {
   rust_source(code = code, env = env, ...)
 }
 
-# Wrapps call to cargo, allowing modification of PATH variable
+#' Sets up environment and invokes Rust's cargo.
+#'
+#' Configures the environment and makes a call to [system2()],
+#'   assuming `cargo` is avaialble on the `PATH`.
+#' Function parameters control the formatting of `cargo` arguments.
+#'
+#' @param toolchain (string) Rust toolchain used for compilation.
+#' @param specific_target (string or `NULL`) Build target (`NULL` if the same as `toolchain`).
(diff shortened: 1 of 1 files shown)


clauswilke:
Could you add a few comments documenting the `invoke_cargo()` function? As is, I don't really understand what the `stdout` and `stderr` arguments do, and that makes is difficulty for me to evaluate the PR.

Ilia-Kosenkov:
@clauswilke, is it better now?

clauswilke:
Yes, looks good to me. Thanks!

PR #116, opened 2021-06-07
Add a vignette of creating an R package

### LINKED ISSUE #74: Write vignette on how to create and update a package with extendr code
We should write a vignette explaining the recommended workflow.

yutannihilation:
Fix #74

Simple version of my blog post, <https://yutani.rbind.io/post/2021-06-06-unofficial-introduction-to-extendr-1-your-first-r-package-with-rust/>. I guess the original one is a bit too casual to turn into the vignette, so I tweaked a bit.

Note that I turned all R chunk to static with pre-executed results pasted. Of course I can create R package on the tempdir and generate the output lively, I'm a bit afraid it becomes a problem on CRAN check.


Ilia-Kosenkov:
Overall this looks good. However, the language needs a little bit of polishing. I'd say, a review by a native speaker or someone living long enough in the English-speaking environment would be good.

I also suggest to make separate PRs for runtime Rust code execution (using `rust_function`, etc.) and for markdown, i.e. what to write in the `.Rmd` to get access to Rust, similar to what is written in the `README.Rmd`.

vignettes/package.Rmd
```diff
@@ -0,0 +1,349 @@
...
+edition = '2018'
+
+[lib]
+crate-type = [ 'staticlib' ]
+
+[dependencies]
+extendr-api = '*'
+```
+
+The create name is the same name as the R package's name by default. You can change this,
```
clauswilke, line 103: "crate"

vignettes/package.Rmd
```diff
@@ -0,0 +1,349 @@
...
+but it might be a bit tired to tweak other files accordingly, so I recommend leaving this.
```
clauswilke, line 104: Replace "tired" with "cumbersome"? Also, "we recommend"

vignettes/package.Rmd
```diff
@@ -0,0 +1,349 @@
...
+and copied to the auto-generated R code. This is analogous to Rcpp/cpp11's `//'`.
+
+``` rs
+/// Return string `"Hello world!"` to R.
+/// @export
+```
+
+The next line is the core of `extendr`'s mechanism. If the function is marked with this
+macro, the corresponding R function will be generated automatically (I'll explain
+the detail later). This is analogous to Rcpp's `[[Rcpp::export]]` and cpp11's `[[cpp11::register]]`.
```
clauswilke, line 148: Remove "I'll epxlain the detail later". This vignette should only be about setting up packages. Other parts of extendr can be explained in other vignettes.

vignettes/package.Rmd
```diff
@@ -0,0 +1,349 @@
...
+    └── rust
+        └── target
+            └── release
+                ├── libmyextendr.a   ---(1)
+                ...
+```
+
+1. **`src/rust/target/release/libmyextendr.a`** (the extension depends on the OS): 
+This is the static library built from Rust code. This will be then used for compiling
+shared library `myextendr.so`.
```
clauswilke, line 232: "the shared library"

vignettes/package.Rmd
```diff
@@ -0,0 +1,349 @@
...
+2. **`src/myextendr.so`** (the extension depends on the OS): This is the
+shared object that is actually called from R.
+3. **`R/extendr-wrappers.R`**: The auto-generated R functions, including roxygen
+comments, goes to this file. The roxygen comments are accordingly converted into
```
clauswilke, line 236: "go into"

vignettes/package.Rmd
```diff
@@ -0,0 +1,349 @@
...
+Rd files and `NAMESPACE`.
+4. **`man/`**, **`NAMESPACE`**: These are generated from roxygen comments.
+
+### Load and use
+
+As all things are done by `rexetndr::document()` already, we can just load it 
+(or install it if you want) and call the function.
```
clauswilke, line 243: Unclear what "it" is here. Write "we can just load the package"?

vignettes/package.Rmd
```diff
@@ -0,0 +1,349 @@
...
+devtools::load_all(".")
+
+hello_world()
+#> [1] "Hello world!"
+```
+
+## Rust code vs generated R code
+
+While we never edit it by hand, it might be good to know what R code is
+generated from a Rust code. Let's look at `R/extendr-wrappers.R`:
```
clauswilke, line 255: "from a Rust code" is weird. Maybe "from some Rust code" or "from specific Rust code"?

vignettes/package.Rmd
```diff
@@ -0,0 +1,349 @@
...
+
+#' Return string `"Hello world!"` to R.
+#' @export
+hello_world <- function() .Call(wrap__hello_world)
+```
+
+`.Call("wrap__make_myextendr_wrappers", use_symbols = ...` is
+what was actually done inside `rextendr::document()`.
+
+A section of `@docType package` is needed to generate `useDynLib(myextendr, .registration = TRUE)` entry
```
clauswilke, line 276: "to generate the ... entry"

vignettes/package.Rmd
```diff
@@ -0,0 +1,349 @@
...
+Now you can call `add()`.
+
+``` r
+devtools::load_all(".")
+
+add(1L, 2L)
+#> [1] 3
+```
+
+## What's next?
```
clauswilke, line 347: Delete from here.

clauswilke (approved):
Looks good to me! I have a number of minor edits to clean up wording/grammar.

vignettes/package.Rmd
yutannihilation: So, should I remove "Rust code vs generated R code" section from this vignette?

vignettes/package.Rmd
yutannihilation: I chose "from the above Rust code." Does this sound fine?

yutannihilation:
Thank you for the thorough review! I reflected the comments.

vignettes/package.Rmd
clauswilke: Yes, good choice.

vignettes/package.Rmd
clauswilke: No, I wouldn't remove it. Just when I read "I'll explain later" it sounded to me as if the vignette was going to give a full-blown intro to how extendr works, and we don't need that here. If you want some sort of a pointer for future material there, maybe just write "... will be generated automatically (see below)". But no pointer at all works also.

clauswilke:
Looks good to me. Please go ahead and merge whenever you're ready.

vignettes/package.Rmd
yutannihilation: Thanks.

vignettes/package.Rmd
yutannihilation: Ah, I see. Let's just remove confusing "I'll explain ...".

yutannihilation:
Thanks, I'll keep this open for a while to proofread this once again, and then merge.

vignettes/package.Rmd
```diff
@@ -0,0 +1,345 @@
...
+---
+
+```{r setup, include=FALSE}
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>"
+)
+```
+
+The rextendr package provides two utility functions for developing R packages with
```
Ilia-Kosenkov, line 17: @yutannihilation,
may I suggest replacing "the rextendr package" with "rextendr" or `rextendr` or any other phrase that does not contain "package"?
There are several examples of such sentences throughout the text, and if we can avoid repetition, it will sound better.

yutannihilation: Oh, for the phrase "the extendr package," I just followed the other vignette. I have no sense which is better.

https://github.com/extendr/rextendr/blob/00ce4650ee044e4b7c66ebeaecaa65c6d5ec7024/vignettes/rmarkdown.Rmd#L15

@clauswilke 
What do you think?

Ilia-Kosenkov: @yutannihilation, I understand that. My point was when you write "the rextendr *package* ... for developing R *packages*", you repeat "package". The closer two "package" usages are, the stranger it sounds. It is perfectly ok to use "The rextendr package", I am just not 100% sure it is the best choice when it is immediately followed by another "package"/"packages" word.


One more example is here

> The package doesn't depend on the rextendr package

It is always challenging to avoid such repetition, but if you can, it is much easier to read, at least for some people, which is probably related to the cultural background/native languages of the reader.

yutannihilation: Ah, thanks. So, maybe it's rather the latter "pacakges" that I need to fix? Or rewrite the whole sentence?

> if you can

I think I can, to some extent, but the problem is it takes me more time to polish the language and I don't have that bandwidth at the moment unfortunately. That's one of the reasons why I originally published this as an "unofficial" one on my personal blog. I'll try to find time to do this, but, if I fail, let me just merge this. You can improve or completely rewrite this after that :)

yutannihilation: Anyway, I really appreciate for pointing out these problems in my writings. Feel free to suggest more drastic rewrites! (though I'm not sure if I can address these in this pull request)

Ilia-Kosenkov: @yutannihilation, sure, I understand, I am no native speaker myself.
If @clauswilke approves, I have no objections. Feel free to modify this PR or make another to improve the wording.

Ilia-Kosenkov (approved):

clauswilke:
I think we shouldn't let minor stylistic issues get in the way of producing useful content. It's good enough as is, and I'd suggest you just go ahead and merge. I can do an additional round of copy-editing afterwards also.

yutannihilation:
Thanks, I'm merging this.

PR #119, opened 2021-06-09
Copy edit development vignette

clauswilke:
I've copy edited the development vignette.

### DIFF
--- a/vignettes/package.Rmd
+++ b/vignettes/package.Rmd
@@ -20,16 +20,13 @@ Rust code using extendr:
 * `rextendr::use_extendr()`: create the scaffolding to use extendr, similar to `usethis::use_cpp11()`.
 * `rextendr::document()`: compile Rust code and generate package documentation, similar to `devtools::document()`.
 
-One thing we want to emphasize here is that these functions are needed solely for
-package development. The package doesn't depend on the rextendr package, just like that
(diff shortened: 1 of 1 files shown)


vignettes/package.Rmd
```diff
@@ -103,7 +100,7 @@ extendr-api = '*'
 The crate name is the same name as the R package's name by default. You can change this,
 but it might be a bit cumbersome to tweak other files accordingly, so we recommend leaving this.
 
-To try the dev version of the extendr, you can modify the last line to
+You will probably want to specify a concrete extendr version, for example `extendr-api = '0.2'`. To try the development version of the extendr, you can modify the last line to read
```
yutannihilation, line 103: I was wondering if I should write this. As long as the package author commits `Cargo.lock`, the build should be reproducible. Of course one might want to specify the older version than the latest released version, but it's probably a rare case.

yutannihilation (approved):
Thanks so much, looks good!

vignettes/package.Rmd
clauswilke: I think it's good practice to specify a version instead of relying only on `Cargo.lock`. Also, I wanted to clarify that the `'*'` isn't there for some important reason.

yutannihilation: I see, sounds good.

PR #125, opened 2021-06-16
Separate compilation and execution of Rust code in `rust_eval()`

### LINKED ISSUE #123: [Feature Proposal]: Separate `rust_eval()` into compilation and evaluation steps for internal use
I propose to modify `rust_eval()` and separate code compilation and import from invocation.
Currently, everything happens within `rust_eval()`:
Compilation
https://github.com/extendr/rextendr/blob/584d67e194b1a8f46745e005e870f0b43f5c9f16/R/eval.R#L41-L49
Execution
https://github.com/extendr/rextendr/blob/584d67e194b1a8f46745e005e870f0b43f5c9f16/R/eval.R#L50
Cleanup
https://github.com/extendr/rextendr/blob/584d67e194b1a8f46745e005e870f0b43f5c9f16/R/eval.R#L51

In order to capture `cargo` output and provide better error messages, we need to be able to inject code in-between compilation and execution. This affects `{knitr}` engines.

My proposed solution is to create an unexported function, which performs compilation and loads library, but instead of evaluation returns a function, which can be called at some later point. This function can be called only once, and upon invocation, it also registers handles for resource cleanup.
If such function is called `rust_eval_deferred`, then the current implementation of `rust_eval()` can be replaced by
```R
rust_eval_deferred <- function(...) { 
  # logic goes here 
  return (function() { })
}
rust_eval <- function(...) {
  rust_eval_deferred(...)()
}
```

This will allow us to separately control how stdout is captured during compilation and during invocation. The former is needed for `cargo` error parsing, the latter -- to be able to knit any `rprintln!()` into the document.

I have a PoC at https://github.com/Ilia-Kosenkov/rextendr/blob/ebeb1ed2e55ae67c9cb1288aa048896f6811bb64/R/eval.R#L46-L104

Ilia-Kosenkov:
Closes #123.

In the current version, `rextendr::rust_eval("some code")` compiles and immediately executes code.
I introduce `rust_eval_deferred()`, an internal function that compiles the code but does not execute it.
Instead, it returns a function that can be called once to execute Rust code and obtain the result. After the first invocation, 
this R function cleans up resources, unloads dll and removes R wrapper function from R environment.

Right now there can be at most one compiled chunk at any time (because once created, it is immediately consumed), so there is no problem with multiple dlls or name overlap. With deferred execution there is a possibility of first compiling several pieces of rust code and then executing them in arbitrary order (though not available through the public API). To address this, I use `rlang::hash()` to compute a unique hash of the input code and dll name (or, rather, a proxy to dll name) and append this hash to the function name. As a result, every time a new code piece is compiled, it gets a unique dll (as was before) and a unique handle on R side. Now this becomes possible:
```R
x <- rust_eval_deferred("10i32") # code compiled and loaded
y <- rust_eval_deferred("20i32") # code compiled and loaded

y() # execute second fragment and free resources
#> 20
x() # execute first fragment and free resources
#> 10
```

This behavior is required for improving `{knitr}` engine and, ultimately, for capturing cargo compilation errors that I am working on. Other potential use cases may emerge in the future.

To emphasize, this **does not** change public API or observed behavior of `{rextendr}`.

I wrote a bunch of tests to verify the execution of multiple deferred chunks in different orders and their resource cleanup.

### DIFF
--- a/R/eval.R
+++ b/R/eval.R
@@ -52,7 +52,16 @@ rust_eval_deferred <- function(code, env = parent.frame(), ...) {
     ui_throw("decoy function; should never be called.")
   }
 
+  # Snippet hash is constructed from the Rust source code and
+  # a unique identifier of the compiled dll.
+  # Every time any rust code is dynamically compiled,
+  # `the$count` is incremented.
+  # This ensures that any two (even bytewise-identical)
+  # Rust source code strings will have different
+  # hashes.
   snippet_hash <- rlang::hash(list(the$count, code))
+
+  # The unique hash is then used to generate unique function names
   fn_name <- glue("rextendr_rust_eval_fun_{snippet_hash}")
 
   # wrap code into Rust function
--- a/R/eval.R
+++ b/R/eval.R
@@ -85,9 +85,8 @@ fn {fn_name}() -> Result<Robj> {{
       )
     }
 
-
-    on.exit(dyn.unload(out[["path"]]), add = TRUE)
-    on.exit(rm(list = fn_name, envir = env), add = TRUE)
+    withr::defer(dyn.unload(out[["path"]]))
+    withr::defer(rm(list = fn_name, envir = env))
 
     result <- rlang::exec(fn_name, .env = env)
     if (has_no_return) {
--- a/R/eval.R
+++ b/R/eval.R
@@ -74,6 +74,11 @@ fn {fn_name}() -> Result<Robj> {{
 }}
 )")
 
+  # Attempt to figure out whether the Rust code returns a result or not,
+  # and make the result invisible or not accordingly. This regex approach
+  # is not perfect, but since it only affects the visibility of the result
+  # that's Ok. Worst case scenario a result that should be invisible is
+  # shown as visible.
   has_no_return <- grepl(".*;\\s*$", code, perl = TRUE)
 
   out <- rust_function(code = code_wrapped, env = env, ...)
--- a/R/eval.R
+++ b/R/eval.R
@@ -29,6 +29,21 @@
 #' }
 #' @export
 rust_eval <- function(code, env = parent.frame(), ...) {
+  rust_eval_deferred(code = code, env = env, ...)()
+}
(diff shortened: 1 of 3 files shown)


R/eval.R
```diff
@@ -37,27 +52,67 @@ rust_eval <- function(code, env = parent.frame(), ...) {
     ui_throw("decoy function; should never be called.")
   }
 
+  snippet_hash <- rlang::hash(list(the$count, code))
```
clauswilke, line 55: Could you add a brief comment here explaining why the hash? You give that explanation in the body of the PR, but I think it's good to record it in the code itself also.

Ilia-Kosenkov: Done

Ilia-Kosenkov:
I added a separate test to verify that two identical source codes produce different handles (different dlls and different wrapper functions), and that both chunks can be safely executed. Basically, we verify hash computations.

R/eval.R
```diff
@@ -37,27 +52,76 @@ rust_eval <- function(code, env = parent.frame(), ...) {
...
-
-  # Attempt to figure out whether the Rust code returns a result or not,
-  # and make the result invisible or not accordingly. This regex approach
-  # is not perfect, but since it only affects the visibility of the result
-  # that's Ok. Worst case scenario a result that should be invisible is
-  # shown as visible.
-  if (grepl(".*;\\s*$", code, perl = TRUE)) {
-    invisible(result)
+
+  has_no_return <- grepl(".*;\\s*$", code, perl = TRUE)
```
clauswilke, line 77: Looks like the comment explaining the grepl line got lost in the shuffle. Could you restore it?

clauswilke (approved):
Looks good to me. One code comment (about guessing whether a return value is produced) got lost in the shuffle. Could you restore it or add something similar?

R/eval.R
Ilia-Kosenkov: Yeah, sure, I indeed lost it.

PR #130, opened 2021-07-17
Support #[extendr(...)]

### LINKED ISSUE #128: Cannot use #[extendr(...)] in rust_source()
As https://github.com/extendr/extendr/pull/222 adds an option to `#[extendr()]` macro, I wonder how the options should be supported on rextendr's side and then wanted to try the code below, but it fails with a cryptic error. I'll take a look later, but please let me know if I'm missing something very obvious...

``` r
code <- r"(
#[extendr(use_try_from = true)]
fn test_i32(val: i32) -> i32 {
    val
}
)"

rextendr::rust_source(code = code, patch.crates_io = list(`extendr-api` = list(git = "https://github.com/extendr/extendr")))
#> ℹ build directory: '/tmp/RtmphmzVDp/file14a9598749e1'
#> Error: Problem with `mutate()` input `..1`.
#> ℹ `..1 = .data$name`.
#> x Column `name` not found in `.data`
```

<sup>Created on 2021-07-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

yutannihilation:
Fix #128

``` r
devtools::load_all("~/repo/rextendr/")
#> ℹ Loading rextendr

code <- r"(
#[extendr(use_try_from = true)]
fn test_i32(val: i32) -> i32 {
    val
}
)"

rust_source(code = code, patch.crates_io = list(`extendr-api` = list(git = "https://github.com/extendr/extendr")))
#> ℹ build directory: '/tmp/RtmpmY1R1g/file15abe30edf0cc'
#> ✓ Writing '/tmp/RtmpmY1R1g/file15abe30edf0cc/target/extendr_wrappers.R'.

test_i32(1)
#> [1] 1
```

<sup>Created on 2021-07-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

### DIFF
--- a/R/find_exports.R
+++ b/R/find_exports.R
@@ -18,7 +18,7 @@ find_exports <- function(clean_lns) {
 
 # Finds lines which contain #[extendr] (allowing additional spaces)
 find_extendr_attrs_ids <- function(lns) {
-  which(stringi::stri_detect_regex(lns, "#\\s*\\[\\s*extendr\\s*\\]"))
+  which(stringi::stri_detect_regex(lns, "#\\s*\\[\\s*extendr(\\s*\\(.+\\))?\\s*\\]"))
 }
 
 # Gets function/module metadata from a subset of lines.
--- a/R/find_exports.R
+++ b/R/find_exports.R
@@ -18,7 +18,7 @@ find_exports <- function(clean_lns) {
 
 # Finds lines which contain #[extendr] (allowing additional spaces)
 find_extendr_attrs_ids <- function(lns) {
-  which(stringi::stri_detect_regex(lns, "#\\s*\\[\\s*extendr(\\s*\\(.+\\))?\\s*\\]"))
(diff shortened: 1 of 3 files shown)


R/find_exports.R
```diff
@@ -18,7 +18,7 @@ find_exports <- function(clean_lns) {
 
 # Finds lines which contain #[extendr] (allowing additional spaces)
 find_extendr_attrs_ids <- function(lns) {
-  which(stringi::stri_detect_regex(lns, "#\\s*\\[\\s*extendr\\s*\\]"))
+  which(stringi::stri_detect_regex(lns, "#\\s*\\[\\s*extendr(\\s*\\(.+\\))?\\s*\\]"))
```
clauswilke, line 21: I find this regex difficult to read. Can we use a raw string? I believe we're requiring R >= 4.0 already anyways.

clauswilke (approved):

R/find_exports.R
yutannihilation: Thanks, let's use a raw string.

PR #140, opened 2021-08-29
Adds support of `features` in `rust_source`

### LINKED ISSUE #139: Support switching `features` during runtime compilation
Currently, `extendr` has at least one feature-guarded...well, feature: `ndarray` support
https://github.com/extendr/extendr/blob/00fed0d8fce2ca0a3c4bf822f685d0d6ada0a1ae/extendr-api/src/prelude.rs#L56-L60

I realized it is currently impossible to write `[features]` to `Cargo.toml`, nor pass any explicit arguments to `cargo` during compilation.

Relevant changes should probably be made here
https://github.com/extendr/rextendr/blob/31c3cd331b0923f17da22c0f27d2168c2014f1e7/R/source.R#L154-L159

Currently `to_toml` supports empty arrays by providing `R` vector of length 0, so enabling `ndarray` feature might look like:

```R

to_toml(...,
    features = list(ndarray = integer(0))
)
```

Alternatively, we can export something like `rextendr::toml_empty_array()`.

Ilia-Kosenkov:
Fixes #139. 
1. Enables `to_toml()` to produce `[ ]` when `NA` value is encountered, e.g. `list(features = list(ndarray = NA))`  translates into
```toml

[features]
ndarray = [ ]

```
2. `rust_source()` gains `features` parameter, defaulting to `NULL`, which, if specified, writes provided values under `features]` section in `Cargo.toml`. Allows enabling `extendr` features like `ndarray` usage.

### DIFF
--- a/R/toml_serialization.R
+++ b/R/toml_serialization.R
@@ -187,13 +187,17 @@ format_toml_atomic <- function(x,
                                ...,
                                .top_level = FALSE,
                                .formatter) {
+  # Cache dimensions because slicing drops attributes
+  dims <- dim(x)
+  x <- x[!is.na(x)]
   len <- length(x)
-  if (len == 0L || (len == 1L && isTRUE(is.na(x)))) {
+  
+  if (len == 0L) {
     "[ ]"
   } else {
     formatter <- as_function(.formatter)
     items <- glue_collapse(formatter(x, ...), ", ")
-    if (len > 1L || !is.null(dim(x))) {
+    if (len > 1L || !is.null(dims)) {
       items <- glue("[ {items} ]")
     }
     # Ensure type-stability
--- a/R/toml_serialization.R
+++ b/R/toml_serialization.R
@@ -187,12 +187,13 @@ format_toml_atomic <- function(x,
                                ...,
                                .top_level = FALSE,
                                .formatter) {
-  if (length(x) == 0L) {
+  len <- length(x)
+  if (len == 0L || (len == 1L && isTRUE(is.na(x)))) {
     "[ ]"
   } else {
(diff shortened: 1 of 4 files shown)


R/toml_serialization.R
```diff
@@ -187,12 +187,13 @@ format_toml_atomic <- function(x,
                                ...,
                                .top_level = FALSE,
                                .formatter) {
-  if (length(x) == 0L) {
+  len <- length(x)
+  if (len == 0L || (len == 1L && isTRUE(is.na(x)))) {
```
clauswilke, line 191: How about stripping all `NA`s from `x` first:
```r
x <- x[!is.na(x)]
```
Then you just need to test whether the length is 0. And I don't think we'd ever want `NA` to get into a toml file anyways.

Ilia-Kosenkov: Great point. I also forgot to update docs...

Ilia-Kosenkov: @clauswilke, 
We rely on `dim()` to determine if provided vector should be displayed as an `array`. 
As usual, base-R slicing drops attributes, so we lose information about dimensions. 
We either need to restore dimensions after slicing, or use something like `vctrs::vec_slice()`, we anyway load it indirectly.
What would be the preferred approach?

Ilia-Kosenkov: Nvm, I cached dimensions before dropping `NA`s, simple & works fine.

clauswilke (approved):

yutannihilation:
Don't we need a NEWS item for this?

clauswilke:
Probably. We have been bad with news items, both here and over at extendr.

PR #149, opened 2021-09-25
Override `CARGO_TARGET_DIR` in `Makevars`

### LINKED ISSUE #145: Override cargo target directory
There are (I think) several ways to specify the directory for artefacts of `cargo` build. One approach is to use [`CARGO_TARGET_DIR`](https://doc.rust-lang.org/cargo/reference/config.html). 

I recently used this var to redirect Rust artefacts from `extendr-api` folder, but then spent half an hour trying to understand why `extendrtests` no longer compile. 

There are two options:
1. `Makevars` should be aware of `CARGO_TARGET_DIR` and use it for linking
2. Supply `--target-dir` to `cargo` in `Makevars` (and when compiling dynamically) to ensure artefacts are emitted correctly.

Ilia-Kosenkov:
Fixes #145.
We hard-code the target directory in `Makevars` (the directory that contains compilation artefacts).
However, if `CARGO_TARGET_DIR` env var is set up, the artefacts are emitted in the subfolder of `CARGO_TARGET_DIR`, and `rextendr::document()` fails to link the library with the `entrypoint.o`. 
This fix checks if `CARGO_TARGET_DIR` is set up and updates relative paths accordingly.

### DIFF
--- a/inst/templates/Makevars
+++ b/inst/templates/Makevars
@@ -1,8 +1,4 @@
-ifeq ($(CARGO_TARGET_DIR),)
 TARGET_DIR = ./rust/target
-else
-TARGET_DIR = $(CARGO_TARGET_DIR)
-endif
 LIBDIR = $(TARGET_DIR)/release
 STATLIB = $(LIBDIR)/lib{{{pkg_name}}}.a
 PKG_LIBS = -L$(LIBDIR) -l{{{pkg_name}}}
@@ -12,7 +8,7 @@ all: C_clean
 $(SHLIB): $(STATLIB)
 
 $(STATLIB):
-	cargo build --lib --release --manifest-path=./rust/Cargo.toml
+	cargo build --lib --release --manifest-path=./rust/Cargo.toml --target-dir $(TARGET_DIR)
 
 C_clean:
 	rm -Rf $(SHLIB) $(STATLIB) $(OBJECTS)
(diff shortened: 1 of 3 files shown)


Ilia-Kosenkov:
Converted to draft because `Makevars` became non-portable. Investigating

yutannihilation:
In my understanding, non-portable `Makevars` is basically fine as long as we write

```
SystemRequirements: GNU make
```

on `DESCRIPTION`. Newer features might not be supported on older platforms (I've seen a few troubles on macOS), but I feel `ifeq` is okay. Actually I'm using some of the GNU make extensions (`:=`, `ifdef`) in [my CRAN package](https://github.com/cran/string2path/blob/031b4bff166999571c16644ee33cef74d3ecdaae/src/Makevars.win) and it's accepted.

yutannihilation:
What I meant by "accepted" is, however, I'm not sure what's the recommended way to customize `Makevars` by environmental variable... We can also use configure scripts to generate `Makevars` from `Makevars.in`.

Ilia-Kosenkov:
I am not entirely convinced my initial approach was correct. Instead of trying to guess what is the preferred cargo target directory on a given system, we should simply override it while invoking `cargo`. We do the same for dynamic calculation. This ensures that we do not emit artifacts in the wrong places and that we have full control over produced binaries.
And it also implifies `Makevars` and `make` extensions are no longer needed.

yutannihilation:
Ah, I got it. Sorry, it seems I didn't read the issue description enough (I thought what we want was some overridable make variable like `TARGET_DIR ?= "./some/default/location/"`). Then this looks good to me!

yutannihilation (approved):

PR #150, opened 2021-10-03
Separate compilation and evaluation in knitr

Ilia-Kosenkov:
The goal of this PR is to replace `rust_eval()` (which was reworked in #125)   with `rust_eval_deferred()`.
With the new function in place, we can separately compile and execute code, allowing greater flexibility for capturing `cargo` output and Rust's `stdout` during execution.

### DIFF
--- a/R/knitr_engine.R
+++ b/R/knitr_engine.R
@@ -47,14 +47,14 @@ eng_impl <- function(options, extendr_engine) {
   }
 
   ui_v("Compiling Rust extendr code chunk...")
-  compilation <- do.call(extendr_engine, c(list(code = code), opts))
+  compiled_code <- do.call(extendr_engine, c(list(code = code), opts))
 
-  if (isTRUE(options$eval) && rlang::is_function(compilation)) {
+  if (isTRUE(options$eval) && rlang::is_function(compiled_code)) {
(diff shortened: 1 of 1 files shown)


Ilia-Kosenkov:
Just to clarify my intention -- I want a more fine-grained control over what information we write to console and what ends up in the knitted file as output. With this approach, we can separately capture outputs of compilation and (where applicable) execution steps.

This PR creates a foundation for this.

R/knitr_engine.R
```diff
@@ -38,16 +38,23 @@ eng_impl <- function(options, rextendr_fun) {
...
   opts <- options$engine.opts
 
-  if (!is.environment(opts$env)) opts$env <- knitr::knit_global() # default env is knit_global()
+  if (!is.environment(opts$env)) {
+    # default env is knit_global()
+    opts$env <- knitr::knit_global()
+  }
+
+  ui_v("Compiling Rust extendr code chunk...")
+  compilation <- do.call(extendr_engine, c(list(code = code), opts))
```
clauswilke, line 50: Maybe call this `compiled` or `compiled_code` or `compiled_rust` instead of `compilation`? To me, `compilation` refers to the process of compiling, not the result of it.

clauswilke (approved):

Issue #100, opened 2021-04-05
Search for `cargo` in a more robust way


malcolmbarrett:
In https://github.com/extendr/libR-sys/issues/54, we ended up discussing whether a `find_cargo()` helper might be useful, similar to blogdown's `find_hugo()`

If I understand @yutannihilation's point, it's that rust, being a whole language, is inherently more complex tooling--including how it was installed--than a single binary like hugo, making this type of solution less obvious for Rust.

My point is that major methods of installation--particularly `rustup`--install Rust in predictable places, and that it's sensible to check those places if `Sys.which("cargo")` doesn't work. For instance, in my case, `cargo` is in `~/.cargo/bin`, right where `rustup` put it.

I was hoping to find the deeper problem in the original thread, but it's still not clear to me why my `PATH` in R doesn't have cargo but I do otherwise, given that my installation and workflow is similar to others. [I seem not to be alone, though](https://stackoverflow.com/questions/66271292/calling-rust-cargo-from-r-results-in-cargo-not-found-error-on-macos/66271293#66271293). Notably, #99 lets me use rextendr successfully, but it's not clear to me how robust it is.

yutannihilation:
Thanks for summarizing.

I'm not totally against the idea, but I'm wondering if it's really "robust" to use `cargo` without having `~/.cargo/bin` in `PATH`. For example, if you want to develop an R package, you'll need `cargo` in `PATH` anyway (c.f. [an example `Makevars`](https://github.com/extendr/helloextendr/blob/24202f849b18e8cd3fe5ec06d9897d3b063ac269/src/Makevars#L10)). So, I think it's good to encourage users to set up their `PATH` correctly instead of secretly solving the problems.

yutannihilation:
> it's still not clear to me why my `PATH` in R doesn't have cargo but I do otherwise (e.g in the terminal), given that my installation and workflow is similar to others

Which do you use zsh or bash? You can check the default shell by `echo $SHELL`. In the case of bash, this post seems a good write up of how setting files such as `~/.profile` and `~/.bashrc` work.

https://scriptingosx.com/2017/04/about-bash_profile-and-bashrc-on-macos/

As I wrote on extendr/libR-sys#54, rustup modifies `~/.profile`, `~/.bashrc` and `~/zshenv`. If I understand correctly,

1. `~/.bashrc` is sourced when you open a new interactive shell, which is not the case when you launch an App from Dock or Finder.
[...]

Issue #99, opened 2021-04-05
Update vignette to suggest `.Renviron` `PATH` solution


malcolmbarrett:
Adding, e.g., `PATH="~/.cargo/bin:${PATH}"` to your `.Renviron` file if rextendr/R can't find `cargo`, likely to here: 
 https://extendr.github.io/rextendr/articles/setting_up_rust.html

I'm filing an issue rather than a PR because I'm actually not sure how robust of a solution it is yet!

clauswilke:
@malcolmbarrett I just checked, on my Mac I have that line in my `.bash_profile` instead of in my `.Renviron`. Have you tested this? Does that not work for you?

malcolmbarrett:
That didn't work for me, nor did modifying other various profile files, including for `zsh`.

What finally works is using `$HOME` rather than `~`. Now my `.Renviron` has:

`PATH="${HOME}/.cargo/bin:${PATH}"`

Now everyone that needs to can find `cargo` from R: 

``` r
system("which cargo")
Sys.which("cargo")
#>                                    cargo 
#> "/Users/malcolmbarrett/.cargo/bin/cargo"
processx::run("which", "cargo")
#> $status
#> [1] 0
#> 
#> $stdout
#> [1] "/Users/malcolmbarrett/.cargo/bin/cargo\n"
#> 
#> $stderr
#> [1] ""
#> 
#> $timeout
#> [1] FALSE
```

TBH, I have no idea why this worked vs. `~`, since it seems like they evaluate to the same value, nor why `Sys.which()` and `processx::run()` required this but `system()` did not

clauswilke:
Interesting! At least this is an easy solution. Also, looking back at my `.bash-profile`, I am actually using `$HOME` instead of `~`. So maybe that's why I've never had any issues.

yutannihilation:
Thanks for the discovery. The expansion of `~` is tricky (e.g. <https://unix.stackexchange.com/a/373532>); it's the feature of the shell, so it needs to be evaluated on the shell, which doesn't automatically happen (spoiler: you'll probably realize this when you want to handle file paths on Rust's side).

If I understand correctly, `system()` simply invokes the shell command, so probably everything is evaluated on shell as usual. On the other hand, processx doing things very programmatically, so I guess it's users' responsibility to expand the `PATH` (it might be considered as a bug, but I'm not sure). No idea about `Sys.which()`, though.

[...]

PR #155, opened 2021-12-12
Use `{processx::run}` to capture `cargo` output

### LINKED ISSUE #115: [Feature Proposal]: Produce better errors when `cargo` fails at runtime
## Before you read
This is an invitation to discussion. There is no need to implement this feature, but I do believe the end users of {extendr} could benefit from something like this.
## Problem outline
We run `cargo` in two scenarios: when a user uses public API like `rust_function` and when {knitr} is processing markdown.
https://github.com/extendr/rextendr/blob/00ce4650ee044e4b7c66ebeaecaa65c6d5ec7024/R/source.R#L281-L297
As you can see, right now we simply use `system2` with some `stderr`/`stdout` redirection and check its exit status.
If not zero, we throw a super-useful error with absolutely **no** details.

In a user session, when R has access to `stdout` and `stderr`, `cargo` displays its output in the console (unless `quiet = TRUE`).
This helps to resolve any compilation errors.
What happens if no `stdout`/`stderr` is available or if `quiet = TRUE`? Well, nothing is printed and the only useful information is the error message saying `"Compilation failed. Aborting"`. This is especially painful when running `R CMD check` or `rcmdcheck::rcmdcheck`, as it shows you errors but not `stdout`/`stderr`. 

## Proposed solution
As suggested by dfalbel in Discord discussion, we use `rlang::abort()` to produce an `rlang::rlang_error` when calling `rextendr::ui_throw()`.
`rlang::abort()` [allows](https://rlang.r-lib.org/reference/abort.html) attaching additional named data to the thrown error using its `...` argument. 
So, we could augment the invocation of `cargo`, capture its errors and attach them to our `rextendr::rextendr_error`. 
This way, even if we disable verbose output for `cargo`, we will be able to provide an explanation to the user of what went wrong.
Sounds simple? Well, not really, here are the details.

### Capturing `cargo` output
The desired solution is to be able to **both** *display* and *capture* output of `cargo`, preserving its color scheme and formatting. We would also want to be able to 'switch off' verbose output yet still capture it for the purpose of error formatting, satisfying `quiet = TRUE`.

`cargo` is a little bit tricky. It prints to `stdout` information like passed/failed tests, but sends all compilation info, warnings and errors to `stderr` (this includes all these fancy `'Updating crates.io`', `'Compiling x'`, `'Finished'`, etc messages).
In our scenario, we run `cargo build --lib`, so there should be no `stdout` at all, but for now let us assume there may be `stdout` during compilation. Just a reminder that we are discussing runtime compilation, where we create Rust crate ourselves, so I expect no `build.rs` or any other build tricks, just plain compilation of (likely one) Rust file(s).

- Problem: [`system2`](https://rdrr.io/r/base/system2.html) cannot simultaneously print and capture output to a variable or file.
- Solution: `cargo` is a separate executable, so let's run it properly: [`processx::run()`](https://rdrr.io/cran/processx/man/run.html). We depend on {processx} through {callr}, which is used for out-of-process wrapper generation. The `processx::run()` allows to both capture and print out `stdout` and `stderr`, separately. If configured correctly, it will behave as `system2()`, but the returned value will contain not only `$status` code, but also `$stdout` and `$stderr`. 
Using additional parameters like `echo` and `echo_cmd`, we can control all of the printed output.
Instead of passing around weird `stdout`/`stderr` variables with obscure values of `""`/`NULL`, we can have one parameter `quiet = logical(1)`, in line with other functions, and switch off all printed output based on the value of this flag.
- Changes: Remove `stdout` and `stderr` from `rextendr:::invoke_cargo()`, add a single `quiet` argument. Replace call to `system2()` with call to `processx::run()` or `callr::run()` (a re-exported function), adjust parameters, record `stderr` in a separate variable.
- Drawbacks:
  -  When capturing `stderr`, on Windows the `stdout`/`stderr` interleaving may break, likely caused by `stdout` buffering.  However, because `cargo` does not write to `stdout` that much in our scenarios, this problem is unobserved. *No remedy is available.*
  - When writing and capturing `stderr`, `processx::run()` prints child process' `stderr` to R's `stdout`. As a result, when using `{knitr}`, compilation information leaks to the captured output. 
  *Temporary fix*: execute {knitr} with `quiet = TRUE`. 
  *Permanent fix*: revise `rextendr::rust_eval()`, which compiles and runs code fragment in one go. Suggested solution: make `rust_eval_fun()` which compiles fragment and returns R wrapper function, which allows to separately capture compilation `stdout` and any `stdout` printed by Rust snippet. Can be implemented together with improving {rextendr} {knitr} engine.
  
- Problem: `cargo` stdout is unstructured. Multiple errors can be emitted alongside warnings. The same stream contains information about successful compilation, which should be excluded when capturing error messages.
- Solution: The output is pretty straightforward. It can be concatenated into a single string. We can match lines on `"\nwarning:|\nerror:"` or something similar. Using this pattern we can split the input into (multi-line) substrings and trim extra spaces from both sides. If matched correctly, then all strings starting with `"^error:"` will contain information about strictly one error. Same applies to `"^warning:"`. 
Collected `errors` and `warnings` can be sent along `rextendr::rextendr_error` as `cargo_errors = list(errors = errors, warnings = warnings)`.
- Drawbacks: 
  - Parsing generally unknown and unstructured output which may change in future versions of Rust. In the worst-case scenario, we won't get any useful information from `stderr`, which is what we have right now (so no regression). In the best-case scenario, we will capture each error and warning separately, ignoring all garbage about successful compilations. This postprocessing only happens when `cargo` fails, so there should be at least one error in the output. Even if we grab everything in the `stderr`, it is still better than what we have right now.
  - {cli} does not support advanced ansi-aware regex, so for now we first strip all of the output of ansi sequences and process plain text. 
  *Permanent fix*: Investigate how to utilize {cli} ansi-aware regex and possibly simplify expressions, which may allow us to include `cargo` error messages in our `rextendr::rextendr_error`, preserving **all** formatting, including colors.
  
### Printing `rextendr_error`
In general, {rlang} does not display additional fields and metadata when printed `rlang::rlang_error`-derived errors. What we would like is to have two display modes: a shorter form which prints only part of errors/warnings, which is used everywhere (especially when the error is uncaught), and a longer form which prints all of the `cargo` errors/warnings, which should be accessed using `summary` `S3` method. 

- Problem: To print `rextendr::rextendr_error` in short form, we need to be able to carefully wrap error messages and subset `n` lines from each error message, preserving format. Otherwise, we will lose useful information like where in the source code the error occurred. It is desirable to offload as much formatting as possible to {rlang} or any other base-type methods.
- Solution:  <s>We can achieve this by overloading two `S3` methods:</s> **This is blocked by** https://github.com/r-lib/rlang/issues/1205
  ```
  conditionMessage.rextendr_error <- function(c) {}
  format.rextendr_error <- function(x, ...) {}
  ```
  `conditionMessage` is invoked by {rlang} when formatting its body. This method provides the output of the base class plus additional information about `cargo` errors in short form. This implementation automatically propagates to other printing methods.
`format` is invoked when formatting, e.g., for `summary` method. We can use one of the {rlang} parameters to determine if the output should be simplified or detailed. When detailed, we output `cargo` errors in the long form. This automatically enables `summary()` to print detailed information.

- Drawbacks: 
  - <s>Due to {rlang}'s [implementation](https://github.com/r-lib/rlang/blob/c2510b8574e4429ccb9eefc805054a611e68f139/R/cnd-error.R) of formatting methods, calling `NextMethod()` in `format()` results in an infninte recursion. </s>
  <s>*Temporary fix*: In `rextendr::format.rextendr_error` temporarily strip error object of all `"rextendr_*"` classes and then dispatch `S3`, which will then resolve into correct {rlang} generics, avoiding infinite recursion. This, however, incorrectly prints the error type.</s>
  *Fix*: Limitations are avoided using options.
  - Wrapping and printing out error messages can be tricky. We want to preserve the structure and (possibly) theme.
    *Temporary fix*: we handle plain text only, wrap lines preserving original line breaks as well.
    *Permanent fix*: we use ansi-aware procedure to process original error messages, preserving full formatting.
    
## The best part: reproducible example:
https://github.com/Ilia-Kosenkov/rextendr/tree/callr-cargo
Successful CI checks of this branch https://github.com/Ilia-Kosenkov/rextendr/actions/runs/838499744
<details>
<summary>Before</summary>

``` r
# Catching error
err <- tryCatch(rextendr::rust_function("fn invalid syntax(){}", quiet = TRUE), error = identity)
# Short form
print(err)
#> <error/rextendr_error>
#> Rust code could not be compiled successfully. Aborting.
#> Backtrace:
#>  1. base::tryCatch(...)
#>  5. rextendr::rust_function("fn invalid syntax(){}", quiet = TRUE)
#>  6. rextendr::rust_source(code = code, env = env, ...)
#>  7. rextendr:::invoke_cargo(...)
#>  8. rextendr:::ui_throw("Rust code could not be compiled successfully. Aborting.")
# Long form
summary(err)
#> <error/rextendr_error>
#> Rust code could not be compiled successfully. Aborting.
#> Backtrace:
#>     x
#>  1. +-base::tryCatch(...)
#>  2. | \-base:::tryCatchList(expr, classes, parentenv, handlers)
#>  3. |   \-base:::tryCatchOne(expr, names, parentenv, handlers[[1L]])
#>  4. |     \-base:::doTryCatch(return(expr), name, parentenv, handler)
#>  5. \-rextendr::rust_function("fn invalid syntax(){}", quiet = TRUE)
#>  6.   \-rextendr::rust_source(code = code, env = env, ...)
#>  7.     \-rextendr:::invoke_cargo(...)
#>  8.       \-rextendr:::ui_throw("Rust code could not be compiled successfully. Aborting.")
```

<sup>Created on 2021-05-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

image

</details>

<details>
<summary>After</summary>

``` r
# Catching error
err <- tryCatch(rextendr::rust_function("fn invalid syntax(){}", quiet = TRUE), error = identity)
# Short form
print(err)
#> <error/rextendr_error>
#> 
#> x `cargo` emitted 3 errors:
#>   1. error: expected one of `(` or `<`, found `syntax`
#>      --> src\lib.rs:3:12
#>       |...
#>   2. error: aborting due to previous error
#>   3. error: could not compile `rextendr1`
#> Backtrace:
#>  1. base::tryCatch(...)
#>  5. rextendr::rust_function("fn invalid syntax(){}", quiet = TRUE)
#>  6. rextendr::rust_source(code = code, env = env, ...)
#>  7. rextendr:::invoke_cargo(...)
#>  8. rextendr:::ui_throw(...)
```

<sup>Created on 2021-05-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

</details>

Ilia-Kosenkov:
This is an attempt to fix #115. Currently, we get little to no information about the reason the compilation has failed.
Now we can throw an error that contains the output of `cargo` marking problematic code fragments.
``` r
rextendr::rust_function("fn wrong_function(x : i33) -> f33 { false }")
#> i build directory: 'C:/Users/[redacted]/AppData/Local/Temp/RtmpM11rtn/file2e184e2156d4'
#> Error: Rust code could not be compiled successfully. Aborting.
#> error[E0412]: cannot find type `i33` in this scope
#>  --> src\lib.rs:3:23
#>   |
#> 3 | fn wrong_function(x : i33) -> f33 { false }
#>   |                       ^^^ help: a builtin type with a similar name exists: `i32`
#> 
#> 
#> error[E0412]: cannot find type `f33` in this scope
#>  --> src\lib.rs:3:31
#>   |
#> 3 | fn wrong_function(x : i33) -> f33 { false }
#>   |                               ^^^ help: a builtin type with a similar name exists: `f32`
#> 
#> 
#> error: aborting due to 2 previous errors
```

<sup>Created on 2021-12-12 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Screenshot 2021-12-12 232131


malcolmbarrett:
Excited about this one!

Ilia-Kosenkov:
@yutannihilation, Windows builds seem to fail because of the missing `-lbcrypt`. Can you make a PR to fix this so I can then merge master and re-run CI?
https://github.com/extendr/extendr/issues/339

R/source.R
```diff
@@ -282,22 +270,63 @@ invoke_cargo <- function(toolchain, specific_target, dir, profile,
...
-    ui_throw("Rust code could not be compiled successfully. Aborting.")
+
+  if (!isTRUE(compilation_result$status == 0)) {
+    errors <-
+      message_buffer %>%
+      purrr::map_chr(stringi::stri_trim) %>%
+      purrr::map(jsonlite::parse_json) %>%
+      purrr::keep(
+        ~.x$reason == "compiler-message" && .x$message$level == "error"
+      )
```
Ilia-Kosenkov, line 317: Here I rely on the fact that `cargo` output is configured to produce JSONs. Should we have an extra `tryCatch` to handle a situation when the output is garbage and can't be parsed as JSON?

Ilia-Kosenkov:
Also, the devel version of R seems to produce weird knitr output, will research it later.
https://github.com/extendr/rextendr/runs/4499559708?check_suite_focus=true#step:11:182

yutannihilation:
Here's the PR: https://github.com/extendr/extendr/issues/339

Ilia-Kosenkov:
So I realized that if an error message is too long (by default it is `options(warning.length = 1000)`, and measured in regular characters, including various invisible ANSI style delimiters), the text gets truncated, and if truncation corrupts ANSI sequence, the terminal gets screwed up (similar to what is shown in the image at the beginning of this thread). The limit can be increased up to I think `8170`, which will be automatically done by `rlang::abort` in the future https://github.com/r-lib/rlang/pull/1214. Still, 8000 is not that much, so first of all I implemented an algorithm that cuts off compiler messages if they no longer fit in the output buffer. The attached image shows how it looks like with a limit of `1000` characters.

image

The second step is to locally override options to set `warning.length` to `8000` so we can display large error messages. This should be enough for the majority of cases, when code fragment has one to five (I guess) reported errors. Same fragment but with a limit of `8000`:

image

The first 'could not compile' error message actually comes from the output of compilation process, and can be silenced with `quiet = TRUE`. Everything else is an error that can be re-inspected with `rlang::last_error()` or `rlang::last_trace()`:

image

Ilia-Kosenkov:
Added `x` to the beginning of each 'error' in the 'details' list.
image

Ilia-Kosenkov:
The changes are significant, I'd like an honest review from as many contributors as possible.

tests/testthat/_snaps/unix/helper-functions.md
```diff
@@ -0,0 +1,25 @@
+# `ui_throw()` called by `rust_function()` captures Rust compilation errors
+
+    Rust code could not be compiled successfully. Aborting.
+    x error[E0412]: cannot find type `i33` in this scope
+     --> src/lib.rs:3:19
```
clauswilke, line 5: Are you sure you want to snapshot these error messages? Any time something minor changes in rustc's output this will break.

R/source.R
```diff
@@ -282,22 +270,65 @@ invoke_cargo <- function(toolchain, specific_target, dir, profile,
...
+      if (!is.null(specific_target)) {
+        sprintf("--target=%s", specific_target)
+      } else {
+        NULL
+      },
+      sprintf("--manifest-path=%s", file.path(dir, "Cargo.toml")),
+      sprintf("--target-dir=%s", file.path(dir, "target")),
+      if (profile == "release") "--release" else NULL,
+      "--message-format=json-diagnostic-rendered-ansi",
+      ifelse(
```
clauswilke, line 294: Why `ifelse` instead of `if () ... else`? `ifelse` is normally used for vectorized data. Also, it's inconsistent with line 292.

clauswilke (commented):
Overall looks good to me. A few minor comments.

tests/testthat/_snaps/unix/helper-functions.md
Ilia-Kosenkov: I understand the danger, yet I trust Rust developer team and believe they will not change the output of `cargo`/`rustc`. Even if they will (e.g., if something like Rust 2.0 is released), I see no problem in updating snapshots. After all, every time `{rlang}`, `{cli}`, `{knitr}`, or any other library that produces formatted output has something changed under the hood, we are forced to update the snapshots (as soon as we notice). This also draws our attention to potential changes that we overlook.

clauswilke: Ok, sounds good.

R/source.R
```diff
@@ -118,16 +118,6 @@ rust_source <- function(file, code = NULL,
   # to be used by `system2()` below
   if (!isTRUE(quiet)) {
     ui_i("build directory: {.file {dir}}")
-
-    # `""` sends `cargo` output to R's standard output.
-    # R console displays to the user informaion about compilation steps and
```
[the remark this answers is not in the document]
Ilia-Kosenkov: I think it was present in the original code fragment, but I deleted it.

R/source.R
Ilia-Kosenkov: Fixed this and a few other arguments. I think `{glue}` works better here than `sprintf()`.

clauswilke:
Any idea what happened with the last series of commits, where now we don't see line-by-line changes anymore for file `R/source.R`?

Ilia-Kosenkov:
Ok, I screwed up line endings, will fix it tomorrow.

Ilia-Kosenkov:
Also, I think we are now missing compilation warnings (like when you have unused variables or parameters). It is not critical, but I'd like to capture that information as well and display it to the user. It should not be difficult, as we already have all the output in JSON. We just need to filter warnings and print them using `{cli}`. I'd like to implement it in this PR.

Ilia-Kosenkov:
So I fixed the issue with line endings, started displaying warnings emitted by cargo, and also fixed a small bug that displayed one line of output even if `quiet = TRUE` was set. Now any interactive compilation is completely silent when `quiet = TRUE` is provided.

Ilia-Kosenkov:
image

yutannihilation:
Looks awesome! I'll try to find time to review this.

Just for information, processx is currently broken on UCRT R-devel on Windows. I'm suspecting the problem is something related to the difference between the OS's encoding (not necessarily UTF-8) and the R's native encoding (UTF-8), so rextendr might be affected. It is probably a bit too early to worry about this (and we don't have CI runner for this), but we might need to have better understanding about the encoding and the stdout/stderr of an external process.

https://cran.r-project.org/web/checks/check_results_processx.html

Ilia-Kosenkov:
When we generate bindings out of process, we use `{callr}`, which AFAIK relies on `{processx}`, so we are already affected. https://github.com/r-lib/callr/blob/ea5c3df7788e43870a265b276d3a4ae103e91557/DESCRIPTION#L20

yutannihilation:
Oh, I see. I think we should add an Windows runner with R-devel to our CI. It seems the UCRT is getting ready.

c.f. https://developer.r-project.org/Blog/public/2021/12/07/upcoming-changes-in-r-4.2-on-windows/index.html

I'll take a look on this.

Ilia-Kosenkov:
Also it would be nice if you can migrate the second CI pipeline to r-actions instead of us doing all the heavy lifting, similar to R CMD check pipeline

R/source.R
```diff
@@ -282,22 +269,117 @@ invoke_cargo <- function(toolchain, specific_target, dir, profile,
...
+        "--color=always"
+      } else {
+        "--color=never"
+      }
     ),
-    stdout = stdout,
-    stderr = stderr
+    echo_cmd = FALSE,
+    windows_verbatim_args = FALSE,
+    stderr = ifelse(isTRUE(quiet), "|", ""),
```
yutannihilation, line 297: You can also replace `ifelse()` with `if` here.

R/source.R
```diff
@@ -282,22 +269,117 @@ invoke_cargo <- function(toolchain, specific_target, dir, profile,
...
+#' and throws errors if compilation was unsuccessful.
+#' @param compilation_result The output of `processx::run()`.
+#' @param message_buffer \[ character(n) \] Messages emitted by cargo to stdout.
+#' @param tty_has_colors \[ logical(1) \] Indicates if output
+#' supports ANSI sequences. If `FALSE`, ANSI sequences are stripped off.
+#' @param quiet Logical indicating whether compile output should be generated or not.
+#' @noRd
+check_cargo_output <- function(compilation_result, message_buffer, tty_has_colors, quiet) {
+  cargo_output <- message_buffer %>%
+    purrr::map_chr(stringi::stri_trim) %>%
```
yutannihilation, line 353: Maybe we can parse JSONs without trimming whitespaces? Just curious.

R/ui.R
```diff
@@ -103,17 +103,66 @@ ui_w <- function(text = "", env = parent.frame()) {
...
+#' @param truncation_notification Message pattern appended if any `lines`
+#' were removed. Understands `cli` interpolation and
+#' `n_removed` integer variable, which is set to the number of removed lines.
+#' @return \[ ansi_character(n) \] A subset of `lines` with
+#' `truncation_notification` appended (if needed).
+#' @noRd
+subset_lines_to_fit_in_limit <- function(lines,
+                                         max_length = 1000,
+                                         truncation_notification = "{.val {n_removed}} compiler messages not shown.") {
+  n_removed <- 100L
```
yutannihilation, line 155:
```suggestion
```

It looks this line is unnecessary.

R/ui.R
```diff
@@ -103,17 +103,66 @@ ui_w <- function(text = "", env = parent.frame()) {
...
+  truncation_notification_size <- nchar(bullet_i(truncation_notification))
```
yutannihilation, line 156: By default, `nchar()` measures the numbers of the characters, but it seems it's actually measured in bytes (see the last one is trancated by 334 repeats).

```
> cat("\u2705")
✅> 
> nchar("\u2705")
[1] 1
> nchar("\u2705", type = "width")
[1] 2
> nchar("\u2705", type = "byte")
[1] 3
> 
> withr::with_options(
+   list(warning.length = 1000),
+   warning(stringr::str_dup("\u2705", 333))
+ )
Warning message:
In force(code) :
  ✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅
> 
> withr::with_options(
+   list(warning.length = 1000),
+   warning(stringr::str_dup("\u2705", 334))
+ )
Warning message:
In force(code) :
  ✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅ [... truncated]
```

yutannihilation (commented):
Generally looks good to me. Minor comments.

R/ui.R
Ilia-Kosenkov: This may be an overkill, but the trick here is to first compute the actual size of the message appended if lines are trimmed. This message is i) customizable ii) depends on the number of removed lines (which is unknown).
So I first assign an arbitrarily large value to `n_removed`, then I use `{cli}` to obtain the actual interpolated string and then use it to compute the length. Assuming we will never have more than 999 lines removed, it gives an upper limit on the length of the added string, which is enough to fit everything within the buffer.
If I remove this declaration, the `{cli}` will fail interpolation.

It is sort of a circular dependency -- `n_removed` depends on the length of the appended message, which in turn depends on `n_removed`.

R/ui.R
Ilia-Kosenkov: Interesting, the documentation on this topic is not as good as I expected, so I assumed we dealt with characters. I will use `type = "byte"`.

R/ui.R
yutannihilation: Ah, sorry, I just missed this variable is referenced by `truncation_notification`. Thanks for the explanation.

yutannihilation (approved):

Ilia-Kosenkov:
I am getting a weird error https://github.com/extendr/rextendr/runs/4592571595?check_suite_focus=true#step:8:368 on Windows, and it does not seem to be reproduced locally. I wonder if it happens here 
```r
 rtools_home <- normalizePath(
      file.path(pkgbuild::rtools_path(), "..", ".."),
      winslash = "/",
      mustWork = TRUE
    )```
and if `pkgbuild::rtools_path()` detects two rtools, `v40` and `v42` for ucrt? Is it possible, @yutannihilation?

Ilia-Kosenkov:
Oh yeah, that's `{pkgbuild}` fault https://github.com/r-lib/pkgbuild/releases/tag/v1.3.1

Ilia-Kosenkov:
https://github.com/r-lib/pkgbuild/commit/2a9ce8d9c25c6dafb29ecb528247f49c10d8c59a#diff-5d9310c23b0c279cdd57b95238eb945cd45bac174c076244ad73cd4147deb626
https://github.com/r-lib/pkgbuild/blob/2a9ce8d9c25c6dafb29ecb528247f49c10d8c59a/R/rtools-metadata.R#L66-L69

R/source.R
Ilia-Kosenkov: You were right, it seems to work fine.

yutannihilation:
Oh, I forgot there's `ucrt64\bin`  bundled in Rtools. Your fix looks good to me.

Btw, we probably need to tweak this to handle the case of `RTOOLS42_HOME`.

https://github.com/extendr/rextendr/blob/9b2f49519b7b86a5b31f8ad20cd652566d43c498/R/source.R#L283

Ilia-Kosenkov:
I think this will be needed for R 4.2.0+. I was hoping you can review and improve this when you investigate running extendr targeting `ucrt` :) I can help you with testing though

yutannihilation:
Yeah, I'll take a look. I'm now using R-devel as my default environment so I'm almost ready for testing it! (except for uninstalling rtools40).

R/ui.R
clauswilke: I would suggest to add a few comments to the code explaining this reasoning. Whenever we have to do some weird trick to get something done, it's a good idea to write the reasoning down, as nobody will remember five years from now.

malcolmbarrett:
I actually can't test this currently, which I think is directly related to #100. I've had trouble with my `PATH` and cargo with processx in the path 🤔

malcolmbarrett:
I wonder if the `env` argument of `processx::run()` can somehow include the local `PATH` if it set up as in #99

Ilia-Kosenkov:
I think I have addressed all the comments. Should I merge?

clauswilke:
No concerns from my side.

malcolmbarrett:
Once you merge this, people in my boat, e.g. who have #99 setup on Macs, may not be able to use the package (as I cannot with this PR). I don't necessarily think it's the responsibility of this PR to address it, but once merged, that issue and #100 become critical.

clauswilke:
Actually, sorry for not connecting the dots earlier. I think @malcolmbarrett is making a good point. If this PR completely bricks the software for people with a reasonable setup (having a path set in `.Renviron` is totally reasonable) we have a problem.

Ilia-Kosenkov:
Ok, I see. I suggest to switch this issue to draft and then investigate processx misbehavior. I have no Apple devices in my household, so it is difficult to debug this issue. I will try to find a MacOS VM to reproduce the problem and find a solution.

yutannihilation:
I'd suggest merging this. As this is the development version, I think it's okay if the `HEAD` gets temporarily broken. As I commented on https://github.com/extendr/rextendr/issues/100#issuecomment-1000564058, I think the only choice is to add `PATH` wherever we use `cargo` (we need to fix the `Makevars` template as well). It probably takes some more changes, not all of which is related to this issue, so I want to separate the discussion, if this sounds okay to @malcolmbarrett.

malcolmbarrett:
Yes, I think that's reasonable. This PR definitely escalates the issue, but I don't think it should necessarily be responsible for addressing it. Your solution looks compelling, @yutannihilation!

yutannihilation:
Thanks for confirming!

Ilia-Kosenkov:
Then I will merge it later today, thanks for the discussion.

PR #156, opened 2021-12-13
Add -lbcrypt

yutannihilation:
Close https://github.com/extendr/extendr/issues/339

As of v1.57.0, Rust requires bcrypt to use `BCryptGenRandom` (Before then, it used to work without bcrypt to support Windows XP).

c.f. https://github.com/rust-lang/rust/pull/84096

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -82,8 +82,6 @@ jobs:
         run: |
           remotes::install_deps(dependencies = TRUE)
           remotes::install_cran("rcmdcheck")
-          #TODO: remove this when testthat is fixed
-          remotes::install_version("testthat", "3.1.0")
         shell: Rscript {0}
 
       - name: Check
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -82,6 +82,8 @@ jobs:
         run: |
           remotes::install_deps(dependencies = TRUE)
           remotes::install_cran("rcmdcheck")
+          #TODO: remove this when testthat is fixed
+          remotes::install_version("testthat", "3.1.0")
         shell: Rscript {0}
 
       - name: Check
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -59,8 +59,8 @@ jobs:
         uses: actions/cache@v2
         with:
           path: ${{ env.R_LIBS_USER }}
-          key: ${{ runner.os }}-${{ hashFiles('.github/R-version') }}-1-${{ hashFiles('.github/depends.Rds') }}
-          restore-keys: ${{ runner.os }}-${{ hashFiles('.github/R-version') }}-1-
+          key: ${{ runner.os }}-${{ hashFiles('.github/R-version') }}-2-${{ hashFiles('.github/depends.Rds') }}
+          restore-keys: ${{ runner.os }}-${{ hashFiles('.github/R-version') }}-2-
 
       - name: Configure Windows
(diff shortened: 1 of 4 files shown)


yutannihilation:
I didn't notice this would affect the snapshot test. Sorry for my laziness...

yutannihilation:
The failure on R-devel ubuntu runner looks unrelated to this change. Probably something related to line breaks is wrong on testthat's side. For example:

```
── Failure (test-knitr-engine.R:27:3): Snapshot test of knitr-engine ───────────
Snapshot of code has changed:
old[6:23] vs new[6:14]
  "  "
  "  "
  "  Basic use example:"
  "  "
+ "  library(rextendr)# create a Rust functionrust_function(\"fn add(a:f64, b:f64) -> f64 { a + b }\")# call it from Radd(2.5, 4.7)#> [1] 7.2"
  "  "
- "  ```r"
- "  library(rextendr)"
  "  "
- "  # create a Rust function"
- "  rust_function(\"fn add(a:f64, b:f64) -> f64 { a + b }\")"
- "  "
- "  # call it from R"
- "  add(2.5, 4.7)"
- "  #> [1] 7.2"
- "  ```"
- "  "
  "  The package also enables a new chunk type for knitr, `extendr`, which compiles and evaluates Rust code. For example, a code chunk such as this one:"
  "  ````markdown"
  "  ```{extendr}"
```

Ilia-Kosenkov:
Should we perhaps fix snapshot test as well?

yutannihilation:
What do you mean? As you can see, nothing is different except for that the result with R-devel is without line breaks.

Actually, it seems the latest version of testthat's snapshot feature is broken somehow, though I still don't understand the details...

https://github.com/r-lib/testthat/issues/1509

yutannihilation:
Hmm, it seems our problem is not related to r-lib/testthat#1509...

Ilia-Kosenkov:
> What do you mean? As you can see, nothing is different except for that the result with R-devel is without line breaks.

Sorry I was reading from my phone and did not realize what the actual problem with the snapshot was.
What would be our strategy? Allow this one to fail for now? I may be able to get a devel version installed locally and see what happens to snapshot tests' output.

yutannihilation:
Ah, I see. Sorry I didn't describe the details in the above comment.

> What would be our strategy? Allow this one to fail for now?

I have no idea at the moment... Let me think.

Ilia-Kosenkov:
@clauswilke, would appreciate your opinion as well.

Ilia-Kosenkov:
Perhaps we can file an issue at `{testthat}`? We have a reproducible case, maybe they can help us?

yutannihilation:
Here's a minimal reproducible example repo. It seems the problem is around `cat_file()`, which we implemented, so it might not be where testthat is responsible for.

https://github.com/yutannihilation/testTestthat

But, really weird thing is, if I switch the GitHub action setting to the latest [`check-pak`](https://github.com/r-lib/actions/blob/master/examples/check-pak.yaml), it won't fail. I have no idea. Both use `rcmdcheck::rcmdcheck()`, so I have no idea why the results are different...

yutannihilation:
I'm still not sure about the root cause of the failure, but it seems the new workflow works so this pull request should be ready to merge. While we can keep investigating what was happening, things on R-devel changes fast, so I think it might not be worth the effort.

Ilia-Kosenkov (approved):
Perhaps we should indeed rely on r-tools to find dependencies and check package instead of doing it ourselves. LGTM, at least for now.

yutannihilation:
I see, thanks. IMHO, if there's now warning using `-lbcrypt` with older Rust, we don't need to do anything, but we'll need to check. I'll file a new issue for this.

PR #166, opened 2021-12-28
Add ~/.cargo/bin to PATH manually

yutannihilation:
Address https://github.com/extendr/rextendr/pull/155#issuecomment-1000565432, fix https://github.com/extendr/rextendr/issues/99

It seems we at last find how to set `PATH` reliably, but, as I commented on #155, there are some CRAN machines that doesn't have `~/.cargo/bin` included (I forgot which, but probably it was macOS). This means, if a package developer aims to release their package on CRAN, they needs to tweak `PATH` anyway. So, while this isn't very cool, it's safe to include `~/.cargo/bin` in `PATH` every time when rextendr invokes `cargo` command.

~~Currently, a test fails if `Makevars` contains `\`-terminated lines. I need to figure out how to fix it~~ (edit: fixed now), but this pull request is basically ready for review.

### DIFF
--- a/R/source.R
+++ b/R/source.R
@@ -278,7 +278,7 @@ invoke_cargo <- function(toolchain, specific_target, dir, profile,
 
   # If there's no `cargo` found, show some helpful messages.
   no_cargo_found <- isTRUE(nchar(Sys.which("cargo")) == 0)
-  if (no_cargo_found) {
+  if (no_cargo_found && !isTRUE(quiet)) {
     ui_w("Can't find {.code cargo} on the PATH. Please review your Rust installation and PATH setups.")
   }
 
@@ -289,8 +289,8 @@ invoke_cargo <- function(toolchain, specific_target, dir, profile,
     # Whether or not `cargo` is found on the PATH, we append `~/.cargo/bin` because
     # it's possible that `rustc` is not found while `cargo` is (c.f. https://github.com/extendr/rextendr/pull/166#discussion_r775803072).
     # So, we use `no_cargo_found` as a condition only for the friendly message.
-    if (no_cargo_found) {
-      ui_i("For now, {.pkg rextendr} will include {.file ${{HOME}}/.cargo/bin} in PATH temporarily. Please consider adding it in {.file ~/.Renviron}.")
+    if (no_cargo_found && !isTRUE(quiet)) {
+      ui_i("{.pkg rextendr} will temporarily include {.file ${{HOME}}/.cargo/bin} in PATH. Please consider adding it to {.file ~/.Renviron}.")
     }
 
     # In some environments, ~/.cargo/bin might not be included in PATH, so we need
--- a/R/source.R
+++ b/R/source.R
@@ -280,9 +280,9 @@ invoke_cargo <- function(toolchain, specific_target, dir, profile,
     # On Windows, PATH to Rust toolchain should be set by the installer
     processx_env <- NULL
   } else {
-    # On Unix-alike, ~/.cargo/bin might not be included in PATH, so we need to
-    # add it here to make sure. It's added to the tail as a fallback, which is
-    # used only when cargo is not found in the user's PATH
+    # In some environments, ~/.cargo/bin might not be included in PATH, so we need
+    # to set it here to ensure cargo can be invoked. It's added to the tail as a
+    # fallback, which is used only when cargo is not found in the user's PATH.
     path_envvar <- Sys.getenv("PATH", unset = "")
     cargo_path <- path.expand("~/.cargo/bin")
     # "current" means appending or overwriting the envvars in addition to the current ones.
--- a/R/source.R
+++ b/R/source.R
@@ -280,11 +280,13 @@ invoke_cargo <- function(toolchain, specific_target, dir, profile,
     # On Windows, PATH to Rust toolchain should be set by the installer
     processx_env <- NULL
   } else {
-  	# On Unix-alike, ~/.cargo/bin is not included in PATH, so we need to set it here to make sure
+    # On Unix-alike, ~/.cargo/bin might not be included in PATH, so we need to
+    # add it here to make sure. It's added to the tail as a fallback, which is
+    # used only when cargo is not found in the user's PATH
     path_envvar <- Sys.getenv("PATH", unset = "")
     cargo_path <- path.expand("~/.cargo/bin")
     # "current" means appending or overwriting the envvars in addition to the current ones.
-    processx_env <- c("current", PATH = glue("{cargo_path}:{path_envvar}"))
+    processx_env <- c("current", PATH = glue("{path_envvar}:{cargo_path}"))
   }
 
   compilation_result <- processx::run(
--- a/R/source.R
+++ b/R/source.R
@@ -283,7 +283,8 @@ invoke_cargo <- function(toolchain, specific_target, dir, profile,
   	# On Unix-alike, ~/.cargo/bin is not included in PATH, so we need to set it here to make sure
     path_envvar <- Sys.getenv("PATH", unset = "")
     cargo_path <- path.expand("~/.cargo/bin")
-    processx_env <- c(PATH = glue("{cargo_path}:{path_envvar}"))
+    # "current" means appending or overwriting the envvars in addition to the current ones.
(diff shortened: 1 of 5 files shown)


yutannihilation:
Hmm, it's really weird. I cannot reproduce the error when I execute the code inside this test step by step...

https://github.com/extendr/rextendr/blob/9413d9790f0743a197ddb021554ff8fae3248a34/tests/testthat/test-use_extendr.R#L45-L78

yutannihilation:
Now I figured out the problem.

```
as.character(glue::glue("a\nb"))
#> [1] "a\nb"
as.character(glue::glue("a\\nb"))
#> [1] "a\\nb"
as.character(glue::glue("a\\\nb"))
#> [1] "ab"
as.character(glue::glue("a\\\\nb"))
#> [1] "a\\\\nb"
```

inst/templates/Makevars
```diff
@@ -8,7 +8,9 @@ all: C_clean
 $(SHLIB): $(STATLIB)
 
 $(STATLIB):
-	cargo build --lib --release --manifest-path=./rust/Cargo.toml --target-dir $(TARGET_DIR)
+	# In some environment, ~/.cargo/bin is not included in PATH, so we need to set it here to make sure
```
Ilia-Kosenkov, line 11: 'In some environment*s*'. Also, it seems the end of the phrase is missing. I suggest 'so we need to set it here to ensure cargo can be invoked'

R/source.R
```diff
@@ -276,6 +276,19 @@ invoke_cargo <- function(toolchain, specific_target, dir, profile,
...
+  } else {
+    # On Unix-alike, ~/.cargo/bin might not be included in PATH, so we need to
+    # add it here to make sure. It's added to the tail as a fallback, which is
+    # used only when cargo is not found in the user's PATH
+    path_envvar <- Sys.getenv("PATH", unset = "")
+    cargo_path <- path.expand("~/.cargo/bin")
+    # "current" means appending or overwriting the envvars in addition to the current ones.
+    processx_env <- c("current", PATH = glue("{path_envvar}:{cargo_path}"))
+  }
+
```
Ilia-Kosenkov, line 291: What if we insert here a check for `cargo` using `Sys.which()`? If I understand correctly, if `Sys.which("cargo")` works, then `processx::run()` also works fine. If that is so, perhaps something like `if (all(nchar(Sys.which("cargo")) == 0)) ui_w("Can't find {.code cargo} on the PATH...")`?

inst/templates/Makevars
yutannihilation: Thanks, I'll use the phrase.

R/source.R
yutannihilation: > if `Sys.which("cargo")` works, then `processx::run()` also works fine.

Unfortunately, no. On CRAN machine, I saw `cargo` is in `PATH` but `rustc` is not (I really don't understand why...).

yutannihilation: Anyway, I guess your point is that if we tweak `PATH`, we should let the user know so that the user can notice the possible problem on their PATH setting, right?

Ilia-Kosenkov: Ok, so does your PR fix this issue on CRAN as well? Or only for regular users?

Ilia-Kosenkov: > Anyway, I guess your point is that if we tweak PATH, we should let the user know so that the user can notice the possible problem on their PATH setting, right?

My point was that if we **fix** the path and **still** cannot find `cargo` or `rustc`, then perhaps it is useful to notify the user, like "hey we did our best but still there is no cargo, your build may fail".

yutannihilation: > does your PR fix this issue on CRAN as well?

Yes. gifski package uses this way: https://github.com/r-rust/gifski/blob/6b86cc6b60abbc2294db821f27cae37413df70c2/src/Makevars#L12-L13

My package sources `~/.cargo/env`, which basically sets `PATH`.

>  notify the user, like "hey we did our best but still there is no cargo, your build may fail".

I see.

Ilia-Kosenkov:
So what happens now is if `cargo` is not on path, every time dynamic compilation happens, a user gets spammed with messages.
This is bad user experience. I was under the impression that there is a package for displaying warnings once per session, but it seems there is none, and `{tidyselect}` uses its own solution https://github.com/r-lib/tidyselect/blob/82d3a2b548d34ceb9e4b337caddfe41e52914986/R/utils.R#L279.

I believe we need to create yet another set of UI helpers to display messages once per session (at least).

R/source.R
Ilia-Kosenkov: I think your warnings do not respect `is_queit`. When it is set to `TRUE`, we should get **no** notifications except for errors thrown with `ui_throw()`.

R/source.R
```diff
@@ -276,10 +276,23 @@ invoke_cargo <- function(toolchain, specific_target, dir, profile,
...
   if (identical(.Platform$OS.type, "windows")) {
     # On Windows, PATH to Rust toolchain should be set by the installer
     cargo_envvars <- NULL
   } else {
+    # Whether or not `cargo` is found on the PATH, we append `~/.cargo/bin` because
+    # it's possible that `rustc` is not found while `cargo` is (c.f. https://github.com/extendr/rextendr/pull/166#discussion_r775803072).
+    # So, we use `no_cargo_found` as a condition only for the friendly message.
+    if (no_cargo_found) {
+      ui_i("For now, {.pkg rextendr} will include {.file ${{HOME}}/.cargo/bin} in PATH temporarily. Please consider adding it in {.file ~/.Renviron}.")
```
Ilia-Kosenkov, line 293: I suggest to leave either 'For now' or 'temporarily', as both of these sentence fragments communicate the same idea. Also, it sounds better if 'Please consider adding it *to*'

R/source.R
yutannihilation: Thanks for catching!

yutannihilation:
> So what happens now is if `cargo` is not on path, every time dynamic compilation happens, a user gets spammed with messages.

I think it's worth spamming if we are to encourage the user to set their `PATH`. It's not the same case as tidyselect. Setting `PATH` properly is not optional. It needs to be done by the user or rextendr package to let things work as expected.

On the other hand, honestly I think we can stop showing the message at all. Considering the discussion on #99, it seems it's not the users' fault but the caveat of R itself (or the trickiness of the shell world), so it might be better to hide the details to avoid confusion.

Ilia-Kosenkov:
Ok, I will support your decision, whether to show messages or not.

yutannihilation:
Thanks, let me think for a while...

yutannihilation:
I made up my mind to choose the latter.

> we can stop showing the message at all

I think we don't need to worry about when our assumption is failed. I bet it should be fairly rare. If we face another issue report on this, let's revisit here...

inst/templates/Makevars
```diff
@@ -8,7 +8,11 @@ all: C_clean
 $(SHLIB): $(STATLIB)
 
 $(STATLIB):
-	cargo build --lib --release --manifest-path=./rust/Cargo.toml --target-dir $(TARGET_DIR)
+	# In some environments, ~/.cargo/bin might not be included in PATH, so we need
+	# to set it here to ensure cargo can be invoked. It's added to the tail as a
```
Ilia-Kosenkov, line 12: I'd propose to rephrase this comment a little bit.
Something like 'It is appended to PATH and therefore is only used (or searched) if cargo is absent from the user's PATH'.
Feel free to change however you like.

tests/testthat/test-source.R
```diff
@@ -56,3 +56,27 @@ test_that("`options` override `rextendr.extendr_deps` value in `rust_source`", {
   withr::local_options(rextendr.extendr_deps = list(`extendr-api` = "-1"))
   expect_rextendr_error(rust_function("fn rust_test() {}"), "Rust code could not be compiled successfully. Aborting.")
 })
+
+test_that("`rust_source` works even when the PATH is not set correctly", {
```
Ilia-Kosenkov, line 60: Should we mention 'MacOS' in the name of the test?

yutannihilation:
Thanks, addressed the comments.

Ilia-Kosenkov (approved):

PR #167, opened 2021-12-28
Changing maintainer of `{rextendr}`

Ilia-Kosenkov:
This PR formalizes the transition of maintainership. 

@clauswilke 
@andy-thomason 
@yutannihilation

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # rextendr (development version)
 
-* Claus Wilke is stepping down as a formal maintainer. Ilia Kosenkov is accepting the role of maintainer.
+* Ilia Kosenkov is now the official maintainer.
 
 * Support extendr macro with options (#128).
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # rextendr (development version)
 
+* Claus Wilke is stepping down as a formal maintainer. Ilia Kosenkov is accepting the role of maintainer.
+
 * Support extendr macro with options (#128).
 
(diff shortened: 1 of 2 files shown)


NEWS.md
```diff
@@ -1,5 +1,7 @@
 # rextendr (development version)
 
+* Claus Wilke is stepping down as a formal maintainer. Ilia Kosenkov is accepting the role of maintainer.
```
yutannihilation, line 3: What about simpler one like "Ilia Kosenkov is now the formal maintainer."?

e.g. https://github.com/r-lib/vroom/blob/main/NEWS.md#vroom-157

yutannihilation (approved):

NEWS.md
clauswilke: Maybe "Ilia Kosenkov is now the official maintainer"?

It should definitely not say "stepping down as *a* formal maintainer", as there can be only one at a time, so it's "*the* formal maintainer".

clauswilke (approved):
Thanks for taking this on, @Ilia-Kosenkov!

NEWS.md
Ilia-Kosenkov: You are probably correct, I did not check how it is usually written in R packages. I'll replace it with what you think is the best option. If `{vroom}` example is OK, I can adopt that.

Ilia-Kosenkov: I used the phrase you both kindly suggested.

yutannihilation:
@Ilia-Kosenkov
I think you can merge this anytime you want.

PR #171, opened 2021-12-30
[PoC] Support R-devel

### LINKED ISSUE #162: Handle RTOOLS42_HOME
https://github.com/extendr/rextendr/pull/155#issuecomment-998791881

yutannihilation:
(Do not merge this, things are changing fast...)

Eventually fixes #162

It turned out Rtools42 is not used yet (ref: https://community.rstudio.com/t/what-is-or-can-be-the-plan-to-support-rtools42-on-r-lib-actions/125114/4) while the R-devel's source code already expects it. Are you wondering like "then, why does this work"? The reason is that Rtools40 also contains the toolchain with the same compiler as Rtools42 under `${RTOOLS40_HOME}/ucrt64`. This pull request tries to handle it.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -45,11 +45,11 @@ jobs:
 
       # Rtools40 contains the GCC10 toolchains in /ucrt64 directory
       - name: Use ucrt64
-        if: startsWith(runner.os, 'Windows') && ${{ matrix.config.r }} == 'devel'
+        if: startsWith(runner.os, 'Windows') && '${{ matrix.config.r }}' == 'devel'
         run: |
           # debug
           echo "${env:RTOOLS40_HOME}"
-          ls -d "${env:RTOOLS40_HOME}/ucrt64"
+          ls "${env:RTOOLS40_HOME}\ucrt64"
           # edd to PATH
           echo "${env:RTOOLS40_HOME}\ucrt64\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
           # confirm
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -45,15 +45,15 @@ jobs:
 
       # Rtools40 contains the GCC10 toolchains in /ucrt64 directory
       - name: Use ucrt64
-        if: startsWith(runner.os, 'Windows') && '${{ matrix.config.r }}' == 'devel'
+        if: startsWith(runner.os, 'Windows') && matrix.config.r == 'devel'
         run: |
           # debug
           echo "${env:RTOOLS40_HOME}"
-          ls "${env:RTOOLS40_HOME}\ucrt64"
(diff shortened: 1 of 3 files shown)


.github/workflows/R-CMD-check.yaml
```diff
@@ -37,11 +38,21 @@ jobs:
           default: true
 
       - name: Configure Windows
-        if: startsWith(runner.os, 'Windows')
+        if: startsWith(runner.os, 'Windows') && matrix.config.r == 'release'
```
Ilia-Kosenkov, line 41: May I suggest to write it as `!= 'devel'`?

.github/workflows/test_pkg_gen.yaml
```diff
@@ -63,14 +64,24 @@ jobs:
           restore-keys: ${{ runner.os }}-${{ hashFiles('.github/R-version') }}-1-
 
       - name: Configure Windows
-        if: startsWith(runner.os, 'Windows')
+        if: startsWith(runner.os, 'Windows') && matrix.config.r == 'release'
```
Ilia-Kosenkov, line 67: Same here

.github/workflows/R-CMD-check.yaml
yutannihilation: Oops, yes... Thanks for catching!

R/source.R
```diff
@@ -248,23 +248,28 @@ invoke_cargo <- function(toolchain, specific_target, dir, profile,
...
-        file.path(
-          rtools_home,
-          paste0("mingw", ifelse(R.version$arch == "i386", "32", "64")),
-          "bin"
-        )
+    if (identical(R.version$crt, "ucrt")) {
+      # TODO: rtools_path() currently returns RTOOLS42_HOME for R >= 4.2, but
+      # current R-devel still uses Rtools40 (c.f. https://community.rstudio.com/t/what-is-or-can-be-the-plan-to-support-rtools42-on-r-lib-actions/125114/4).
+      # So, we need a tweak for now. Eventually, this can be removed.
+      rtools_home <- Sys.getenv("RTOOLS40_HOME", "C:\\rtools42")
```
Ilia-Kosenkov, line 255: Out of curiosity, why can't we use here `pkgbuild::rtools_path()`? Their latest release has some support for two versions of Rtools on the system.

yutannihilation: `pkgbuild::rtools_path()` always returns the path to Rtools42, which isn't used yet.

https://github.com/r-lib/pkgbuild/blob/161f7d31d23da1c89fe9959679a6fd3150eba515/R/rtools.R#L34

> Their latest release has some support for two versions of Rtools on the system.

It's a bit confusing, but what they added was a support for Rtools40v2, which contains both toolchain for R 4.1 (`/mingw*`) and that for R 4.2 (`/ucrt64`).

Ilia-Kosenkov: Ok, I see.

yutannihilation:
This was written for Rtools40. Closing this for now. I'll create a new one soon...

PR #174, opened 2022-01-13
Update GitHub Actions

yutannihilation:
* Use v2
* Update the standard GitHub Actions with the latest ones on r-lib/actions

### DIFF
--- a/.github/workflows/pkgdown.yaml
+++ b/.github/workflows/pkgdown.yaml
@@ -1,22 +1,25 @@
+# Workflow derived from https://github.com/r-lib/actions/tree/v2/examples
+# Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
 on:
   push:
-    branches:
-      - main
-      - master
+    branches: [main, master]
+  pull_request:
+    branches: [main, master]
+  release:
+    types: [published]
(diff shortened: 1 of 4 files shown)


.github/workflows/pkgdown.yaml
```diff
@@ -1,54 +1,48 @@
...
   push:
-    branches:
-      - main
-      - master
+    branches: [main, master]
+  pull_request:
+    branches: [main, master]
+  release:
+    types: [published]
+  workflow_dispatch:
```
Ilia-Kosenkov, line 10: What is the reason for changing triggers? Is it because of the new template? Do we really need anything other than 'on push into master/main'?

yutannihilation: I didn't notice this. Thanks for catching.

It seems the current strategy is build all the time and deploy conditionally. I think this is a good idea because we can notice the pkgdown failures (e.g. missing topics) before actually merging the pull request. But, if we want to save our CI time, I'll remove this.

``` yaml
      - name: Deploy to GitHub pages 🚀
        if: github.event_name != 'pull_request'
        uses: JamesIves/github-pages-deploy-action@4.1.4
```

Ilia-Kosenkov (approved):

PR #180, opened 2022-02-14
Allow different crate and library names, R pacakges with dots in their names

### LINKED ISSUE #152: Develop R package with name not equal to `package.name` in `Cargo.toml`
I have a rust crate called `hdcd` and would like to build an API for R using `rextendr`. The resulting R package should also be called `hdcd`. For this I create a new rust package `hdcdr`, which wraps `hdcd` and implements some `From<...> for Robj` traits. Note that this wrapping crate cannot have `hdcd` as a package name, as it has `hdcd` as a dependency.

I have been unable to configure the package such that the R-package name (should be `hdcd`) is different from the `package.name` entry in `Cargo.toml`. If I just set the `Package: ` entry in the `DESCRIPTION` to `hdcd`, I get the following error upon `rextendr::document()`:

```
Error in .Call(wrapper_function, use_symbols = use_symbols, package_name = package_name,  : 
  "wrap__make_hdcd_wrappers" not available for .Call() for package "hdcd"
Error: Failed to generate wrapper functions.
✖ callr subprocess failed: "wrap__make_hdcd_wrappers" not available for .Call() for package "hdcd"
```
Here is the equivalent issue for pyo3: https://github.com/PyO3/maturin/issues/313#issuecomment-784438527. 

Following the advice there, I added a `lib.name = 'hdcd'` entry to `Cargo.toml`. Then, I get the following error:
```
   x86_64-conda-linux-gnu-cc -shared -L/home/mlondschien/miniforge3/envs/rust/lib/R/lib -Wl,-O2 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,--disable-new-dtags -Wl,--gc-sections -Wl,-rpath,/home/mlondschien/miniforge3/envs/rust/lib -Wl,-rpath-link,/home/mlondschien/miniforge3/envs/rust/lib -L/home/mlondschien/miniforge3/envs/rust/lib -o hdcd.so entrypoint.o -L./rust/target/release -lhdcdr -L/home/mlondschien/miniforge3/envs/rust/lib/R/lib -lR
   /home/mlondschien/miniforge3/envs/rust/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/../../../../x86_64-conda-linux-gnu/bin/ld: cannot find -lhdcdr
   collect2: error: ld returned 1 exit status
   make: *** [/home/mlondschien/miniforge3/envs/rust/lib/R/share/make/shlib.mk:10: hdcd.so] Error 1
   ERROR: compilation failed for package ‘hdcd’
```

I feel like this might be fixed by monkeypatching the `PKG_LIBS` entry in the `Makevars` file from `PKG_LIBS = -L$(LIBDIR) -lhdcdr` to `PKG_LIBS = -L$(LIBDIR) -lhdcd`. Doing so I get back to the first error: 
```
Error in .Call(wrapper_function, use_symbols = use_symbols, package_name = package_name,  : 
  "wrap__make_hdcd_wrappers" not available for .Call() for package "hdcd"
Error: Failed to generate wrapper functions.
✖ callr subprocess failed: "wrap__make_hdcd_wrappers" not available for .Call() for package "hdcd"
```
Any suggestions on how to proceed?

### LINKED ISSUE #176: Packages with `.` in name do not compile
The `.` part of the package name results in invalid C and Rust code, it seems

``` r
library(usethis)
library(rextendr)
path <- file.path(tempdir(), "my.pkg")
ui_silence(create_package(path))
setwd(path)
rextendr::use_extendr()
#> ✓ Creating 'src/rust/src'.
#> ✓ Setting active project to '/private/var/folders/w7/8yv1j00s0bb3pfhmqc_rvd980000gn/T/RtmpTYmcmp/my.pkg'
#> ✓ Writing 'src/entrypoint.c'
#> ✓ Writing 'src/Makevars'
#> ✓ Writing 'src/Makevars.win'
#> ✓ Writing 'src/.gitignore'
#> ✓ Writing 'src/rust/Cargo.toml'.
#> ✓ Writing 'src/rust/src/lib.rs'
#> ✓ Writing 'R/extendr-wrappers.R'
#> ✓ Finished configuring extendr for package my.pkg.
#> • Please update the system requirement in 'DESCRIPTION' file.
#> • Please run `rextendr::document()` for changes to take effect.
rextendr::document()
#> ℹ Generating extendr wrapper functions for package: my.pkg.
#> ! No library found at 'src/my.pkg.so', recompilation is required.
#> Re-compiling my.pkg
#> * installing *source* package ‘my.pkg’ ...
#> ** using staged installation
#> ** libs
#> rm -Rf my.pkg.so ./rust/target/release/libmy.pkg.a entrypoint.o
#> clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include   -fPIC  -falign-functions=64 -Wall -g -O2  -UNDEBUG -Wall -pedantic -g -O0 -c entrypoint.c -o entrypoint.o
#> entrypoint.c:4:6: error: variable has incomplete type 'void'
#> void R_init_my.pkg_extendr(void *dll);
#>      ^
#> entrypoint.c:4:15: error: expected ';' after top level declarator
#> void R_init_my.pkg_extendr(void *dll);
#>               ^
#>               ;
#> entrypoint.c:6:6: error: variable has incomplete type 'void'
#> void R_init_my.pkg(void *dll) {
#>      ^
#> entrypoint.c:6:15: error: expected ';' after top level declarator
#> void R_init_my.pkg(void *dll) {
#>               ^
#>               ;
#> 4 errors generated.
#> make: *** [entrypoint.o] Error 1
#> ERROR: compilation failed for package ‘my.pkg’
#> * removing ‘/private/var/folders/w7/8yv1j00s0bb3pfhmqc_rvd980000gn/T/RtmpTYmcmp/devtools_install_108de4e22d649/my.pkg’
#> Error in (function (command = NULL, args = character(), error_on_status = TRUE, : System command 'R' failed, exit status: 1, stdout & stderr were printed
```

<sup>Created on 2022-02-08 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Ilia-Kosenkov:
- Fixes #176, allowing R package names like `{test.package}`
- Fixes #152, allowing to set `crate name` and `library name` (= the name of the binary produced by Rust) separately from the name of the R package, resolving potential crate/package name collisions.

'Names' we distinguish:
- ✔ R package name, sanitized if necessary. This is the name of the final binary artifact and the name of the topmost module defined in `src/rust/lib.rs`;
- ✔ Rust crate name, which defaults to the sanitized R package name. May be helpful in resolving conflicts with cargo dependencies;
- ✔ Rust library name, which defaults to the sanitized R package name. Same usage as described above;
- ❌  <s>We can potentially distinguish between R package name and topmost exported Rust module, but this may need some support from `extendr` </s> If the module name differs from that of the R package, we will have to pass this name to `{rextendr}` every time we call `rextendr::document()`, because this name determines the name of the function that produces wrappers. I suggest to not support this feature.

TODO:
- [x] Check docs
- [x] Run {styler}
- [x] Write tests for package/crate name combinations


Ilia-Kosenkov:
I think I covered all cases.

R/use_extendr.R
```diff
@@ -108,6 +128,44 @@ use_extendr <- function(path = ".", quiet = getOption("usethis.quiet", FALSE)) {
...
+#' Checks if provided name is a valid Rust name (identifier)
+#'
+#' @param name \[ character(n) \] Names to test.
+#' @return \[ logical(n) \] `TRUE` if the name is valid, otherwise `FALSE`.
+#' @noRd
+is_valid_rust_name <- function(name) {
+  # We require the name starts with a letter,
+  # ends with a letter or digit,
+  # and contains only alphanumeric ASCII chars, `-` or `_`.
+  stringi::stri_detect_regex(name, "^[A-z][\\w-]*[A-z0-9]$")
```
yutannihilation, line 140: Probably we don't get any non-ASCII characters here, so this is just a nitpicking, but I think `\w` is too wide. On ICU, it means `[\p{Alphabetic}\p{Mark}\p{Decimal_Number}\p{Connector_Punctuation}\u200c\u200d]` ([ref](https://unicode-org.github.io/icu/userguide/strings/regexp.html)) and `\p{Alphabetic}` includes non-ASCII characters. 

``` r
stringi::stri_detect_regex("錆", "\\w")
#> [1] TRUE
```

<sup>Created on 2022-02-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

R/use_extendr.R
```diff
@@ -108,6 +128,44 @@ use_extendr <- function(path = ".", quiet = getOption("usethis.quiet", FALSE)) {
...
+}
+
+#' Verifies if a function argument is a valid Rust name.
+#'
+#' @param name \[ string \] Tested caller function argument.
+#' @param call \[ env \] Environment of the caller, passed to `ui_throw()`.
+#' @noRd
+throw_if_invalid_rust_name <- function(name, call = caller_env()) {
+  quo <- enquo(name)
+  if (!is.character(name) || length(name) != 1 || !is_valid_rust_name(name)) {
```
yutannihilation, line 161:
```suggestion
  if (length(name) != 1 || !is.character(name) || !is_valid_rust_name(name)) {
```

`is.character()` might return length >1 vector when `name` is length >1.

yutannihilation (approved):
Looks good

R/use_extendr.R
Ilia-Kosenkov: I hate base R conventions so much sometimes. We should probably explicitly reference `{vctrs}` in the future and use it for checks.

yutannihilation: Ah, I just remembered we can simply use `rlang::is_scalar_character()`.

Ilia-Kosenkov:
I think everything has been resolved. I am going to merge this if there are no objections.

yutannihilation:
Looks good, in case you need another explicit approval!

PR #181, opened 2022-02-14
[docs] Link to docs on creating a Rust + R package

sstadick:
Hello! This is just a small PR to point people in the right direction for creating a mixed R / Rust library, which I had a hard time finding when looking at the READMEs across extendr projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,9 @@ z
 #> [1] 35
 ```
 
+To create library based on Rust source code, see the following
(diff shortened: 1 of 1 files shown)


yutannihilation (approved):
Thanks! I feel, ideally, this document should be on <https://extendr.github.io/>, but I don't think I can have time to improve the documents there in the near future, so this link should be good for now.

PR #187, opened 2022-03-27
Use rtools42

yutannihilation:
This pull request is the successor of #171.

Note that this pull request assumes we use `stable-gnu` on R-devel because it's difficult to support both, but we don't decide which to use `stable-gnu` or `stable-msvc`. If we choose the latter, I'll rewrite the corresponding codes.

### DIFF
--- a/R/source.R
+++ b/R/source.R
@@ -281,7 +281,18 @@ invoke_cargo <- function(toolchain, specific_target, dir, profile,
     # On Windows, PATH to Rust toolchain should be set by the installer.
     # If R >= 4.2, we need to override the linker setting.
     if (identical(R.version$crt, "ucrt")) {
-      cargo_envvars <- c("current", CARGO_TARGET_X86_64_PC_WINDOWS_GNU_LINKER = "x86_64-w64-mingw32.static.posix-gcc.exe")
+      # libgcc_mock is created in configure.ucrt on installation.
+      libgcc_path <- system.file("libgcc_mock", package = "rextendr")
+      if (identical(libgcc_path, "")) {
+        ui_throw(
+          "Unable to find {.file inst/libgcc_mock}. Please reinstall the rextendr package",
+        )
+      }
+
+      cargo_envvars <- c("current",
+        CARGO_TARGET_X86_64_PC_WINDOWS_GNU_LINKER = "x86_64-w64-mingw32.static.posix-gcc.exe",
+        LIBRARY_PATH = paste0(libgcc_path, ";", Sys.getenv("LIBRARY_PATH"))
+      )
     } else {
       cargo_envvars <- NULL
     }
--- a/R/source.R
+++ b/R/source.R
@@ -251,24 +251,25 @@ invoke_cargo <- function(toolchain, specific_target, dir, profile,
     # rtools_path() returns path to the RTOOLS40_HOME\usr\bin,
     # but we need RTOOLS40_HOME\mingw{arch}\bin, hence the "../.."
     rtools_home <- normalizePath(
-      # `pkgbuild` may return two paths for R < 4.2 if Rtools42 is present
+      # `pkgbuild` may return two paths for R < 4.2 with Rtools40v2
       file.path(pkgbuild::rtools_path()[1], "..", ".."),
       winslash = "/",
       mustWork = TRUE
     )
 
-    rtools_bin_path <-
-      normalizePath(
-        file.path(
-          rtools_home,
-          paste0("mingw", ifelse(R.version$arch == "i386", "32", "64")),
-          "bin"
-        )
-      )
-    # Appends path to rtools\mingw{arch}\bin using a correct arch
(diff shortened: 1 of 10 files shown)


yutannihilation:
No idea why only the test of dot name fail in this way. I cannot reproduce this issue on my local. This needs some investigation...


`E>           x86_64-w64-mingw32.static.posix-gcc.exe: error: C:\Users\runneradmin\AppData\Local\Temp\RtmpGIYi8A\working_dir\Rtmpa4uQZH\file6e0254a6d7\package.with.dots\src\./rust/target\release\build\winapi-x86_64-pc-windows-gnu-4bcab0af3bd55335\build_script_build-4bcab0af3bd55335.build_script_build.f5023671-cgu.9.rcgu.o: No such file or directory`
`E>           `
`E> `
``E> error: could not compile `winapi-x86_64-pc-windows-gnu` due to previous error``

R/source.R
```diff
@@ -251,24 +251,25 @@ invoke_cargo <- function(toolchain, specific_target, dir, profile,
     # rtools_path() returns path to the RTOOLS40_HOME\usr\bin,
     # but we need RTOOLS40_HOME\mingw{arch}\bin, hence the "../.."
     rtools_home <- normalizePath(
-      # `pkgbuild` may return two paths for R < 4.2 if Rtools42 is present
+      # `pkgbuild` may return two paths for R < 4.2 with Rtools40v2
       file.path(pkgbuild::rtools_path()[1], "..", ".."),
       winslash = "/",
       mustWork = TRUE
     )
```
Ilia-Kosenkov, line 258: Are we sure that the first returned path is the correct one? What if both v40 and v42 are installed on the system?

yutannihilation: Yes, I'm pretty sure.

https://github.com/r-lib/pkgbuild/blob/a22b6a3014476efba60adae77858d4ff7ebc1deb/R/rtools.R#L37-L43

Ilia-Kosenkov:
I will try to look into these errors later today. Need to first configure R-devel & ucrt toolchain locally. Indeed it is strange, as a hundred of other tests succeed.

mati865:
> No idea why only the test of dot name fail in this way. I cannot reproduce this issue on my local. This needs some investigation...
> 
> `E> x86_64-w64-mingw32.static.posix-gcc.exe: error: C:\Users\runneradmin\AppData\Local\Temp\RtmpGIYi8A\working_dir\Rtmpa4uQZH\file6e0254a6d7\package.with.dots\src\./rust/target\release\build\winapi-x86_64-pc-windows-gnu-4bcab0af3bd55335\build_script_build-4bcab0af3bd55335.build_script_build.f5023671-cgu.9.rcgu.o: No such file or directory` `E> ` `E> ` `` E> error: could not compile `winapi-x86_64-pc-windows-gnu` due to previous error ``

I believe this hits Windows PATH limit since GCC/Binutils do not add manifest allowing to use longer than default paths.

Ilia-Kosenkov:
> I believe this hits Windows path length limit since GCC/Binutils do not add manifest allowing to use longer than default paths.

This is an interesting idea. The path you quote seems to be 262 characters long, while Windows (by default) supports only [260](https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=cmd) characters. It is possible that developers like myself or @yutannihilation have workstations configured to use longer paths and therefore not observe this error. We can test it by changing that name to, say, `pkg.w.dts`.

yutannihilation:
Oh, thanks. It can explain the problem. Let's try some shorter package name.

yutannihilation:
🎉

Ilia-Kosenkov:
We definitely need to re-consider temporary cargo build directories to avoid similar problems in the future.
Great job by the way!

yutannihilation:
Thanks! I'm yet to understand what the problem is, but I guess using a temp dir is a trade-off between the compilation speed. Anyway let's discuss in another issue.

Thanks @mati865 for your great insight. It really helped us!

yutannihilation:
What I tried so far:

* From the documents, we can use `RUSTFLAGS` envvar to inject `-L path/to/lib_dir` argument to `rustc`, but it seems this won't work for all of the dependency crates. I couldn't figure out the rule here.
* Instead, it seems we can use `LIBRARY_PATH` envvar, which probably is referred from the GCC toolchain directly.
* I'm yet to figure out how to generate the empty libgcc_{eh,s} in `cargo_invoke()`.
* While this approach is good for R packages, I have no idea what we should do for non-R packages.

yutannihilation:
This pull request is ready for review now, but probably the description is too terse now. I'll write more, but my priority is https://github.com/extendr/libR-sys/pull/94 for now. But feel free to ask questions...

Ilia-Kosenkov (approved):
Yep, please go ahead.

yutannihilation:
Thanks! If you find any problem when using rextendr with R 4.2, please let me know.

PR #193, opened 2022-05-11
Fix document

CGMossa:
This is due to `||` causing an error with more than 1 conditions in R 4.2.

### DIFF
--- a/R/track_rust_source.R
+++ b/R/track_rust_source.R
@@ -49,11 +49,11 @@ pretty_rel_path <- function(path, search_from = ".") {
   # return `path` unchanged (for simplicity).
   if (
     !nzchar(package_root) ||
-      !stringi::stri_detect_fixed(
+      !all(stringi::stri_detect_fixed(
         str = path,
         pattern = package_root,
         case_insensitive = TRUE
-      )
+      ))
   ) {
     return(path)
   }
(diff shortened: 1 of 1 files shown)


R/track_rust_source.R
```diff
@@ -48,11 +49,11 @@ pretty_rel_path <- function(path, search_from = ".") {
   # return `path` unchanged (for simplicity).
   if (
     !nzchar(package_root) ||
-      !stringi::stri_detect_fixed(
+      !all(stringi::stri_detect_fixed(
```
Ilia-Kosenkov, line 52: This is not exactly the expected behavior. I'd suggest to keep this function as is, but rename it. And then introduce the new `pretty_rel_path()` as 
```R
pretty_rel_path <- function(path, search_from = ".") {
   purrr::map_chr(path, pretty_rel_path_impl, search_from = search_from)
}
```
to add proper vectorization support.

CGMossa:
I think this is weird to make a test out of, but what do you think of this:

```r
test_that("Running `document` after adding multiple files", {
  skip_on_cran()

  path <- local_package("testPackage")
  rextendr::use_extendr()
  rextendr::document()

  file.create(glue("touch {path}src/rust/src/a.rs"))
  file.create(glue("touch {path}src/rust/src/b.rs"))

  rextendr::document()
})

```

Ilia-Kosenkov:
Shouldn't we have an `expect_*()` call for `{testthat}` to work properly?

CGMossa:
Yeah, but what should I expect? If an error happens then the test fails. The files are empty so there is no snapshot to be made and compared..

Ilia-Kosenkov:
According to the `{testthat}` [docs](https://testthat.r-lib.org/reference/expect_error.html) (see the description of parameter `regexp`), one can write `exepct_error(code(), NA)` to signal that there should be **no** error.

Ilia-Kosenkov (approved):

PR #208, opened 2022-09-26
Allow controlling Rust toolchain via TOOLCHAIN env variable

Ilia-Kosenkov:
This trivial modification of the `Makevars.ucrt` template allows propagating `TOOLCHAIN` environment variable to the package build step. While on platforms other than Windows default Rust toolchain controls how the package is built, the toolchain was hardcoded for Windows in the makevars files.

One of the usecases is an easy switch to `nightly` version of the toolchain, see e.g. https://github.com/extendr/helloextendr/pull/15.


yutannihilation (approved):
Sorry for the delay. I personally prefer using the `configure` script to inject different settings, but this also looks good to me.

PR #212, opened 2022-11-14
Limit dll exports on windows

### LINKED ISSUE #211: `x86_64-w64-mingw32.static.posix\bin/ld.exe` linker cannot handle large rust crates
## problem:
The following error occours if compiling a big project with latest rextendr on windows:
`C:\rtools42\x86_64-w64-mingw32.static.posix\bin/ld.exe: error: export ordinal too large: 88000`

## issue
It seems the rtools42 gnu linker currently uses a [16bit enumeration of symbols (ordinal symbol table)](https://stackoverflow.com/questions/47135973/error-export-ordinal-too-large-104116). If using a large rust crate such as `polars` the compilation will fail for windows as this table is exceeded.

## Minimal reproducible example:
[I have made a minimal inclusion of `polars`in helloextendr to show the error here](https://github.com/sorhawell/helloextendr_big_windows/actions/runs/3413285972/jobs/5679824782)

## What I could google:
[one tried to solve the issue by moving back `mvsc` another needed to use `gnu/msys2`  but alledgely got by doing some mix msys2+mvsc compiling](https://users.rust-lang.org/t/export-ordinal-too-large-error-while-linking/74547/7)

[Some golang people have the same isssue and are discussion to limit the number of exported symbols](https://github.com/golang/go/issues/40795)

... and it seems the gophers found some kind of [ solution here](https://github.com/golang/go/issues/30674)

Ilia-Kosenkov:
Closes #211. 
When scaffolding a package to use with `extendr`, also emit a `<packagename>-win.def` file with exports to avoid exporting too many symbols on Windows.


CGMossa (approved):
Lgtm

PR #213, opened 2022-11-20
add CRAN install instruction, relabel remotes install command as dev …

bquast:
…version install

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,12 @@ stable](https://img.shields.io/badge/lifecycle-stable-brightgreen.svg)](https://
 To install the package, run:
 
 ``` r
+install.packages("rextendr")
+```
+
+To install the development version run:
+
+``` r
 remotes::install_github("extendr/rextendr")
 ```
 


Ilia-Kosenkov:
The CRAN version is outdated, I am not sure it is even useful to install it from CRAN. It seems using r-universe would be much more preferrable.

bquast:
I see, but going forward, wouldn't it be good to have a stable on CRAN and a dev on GitHub?

I think this is a very important package, a reasonable recent, stable version on CRAN is important right?

If it helps I can contribute to maintaining.

CGMossa:
There are reasons beyond our reach for the lack of an updated CRAN version.

Ilia-Kosenkov:
> If it helps I can contribute to maintaining.

We welcome contributions and fresh ideas!

CRAN situation is difficult. While we can push `{rextendr}` (there is a 0.3.0 release [planned](https://github.com/extendr/rextendr/milestone/2)), CRAN is hostile towards `extendr`-consuming packages, or any packages using Rust, and we have not found a good way of dealing with it so far.

yutannihilation:
Just in case you don't notice, since the README is generated from .Rmd, we have to edit and render this one.

https://github.com/extendr/rextendr/blob/main/README.Rmd

bquast:
That's a shame about the hostility, I would argue that with PETs becoming prominent, a safe language like Rust is also relevant to the R ecosystem.

Have you asked Dirk Eidelbuettel of Rcpp? maybe he could offer advice?

Yes, sorry about the Rmd, I did notice, and then I forgot. But it also seems atm maybe a change is not needed yet.

Ilia-Kosenkov:
Let's close this for now since we have not arrived at a decision which way to go -- full CRAN or full R Universe

eitsupi:
Could you consider a new CRAN release?
With modification by #233, packages using the rextendr template would be accepted by CRAN.
[prqlr on CRAN](https://cran.r-project.org/package=prqlr) uses the template from #233 and there are no other hacks.

Issue #169, opened 2021-12-30
Integration with `{lintr}`


Ilia-Kosenkov:
[`{lintr}`](https://github.com/r-lib/lintr) is a package that highlights issues in the codebase. It can be used as a standalone tool or as a part of an editor/IDE (such as vscode). I propose to research `{lintr}` application to `{rextendr}` to improve the quality of code.

- [ ] Configure `{lintr}` using `.lintr` file to enforce rules
- [ ] Add `{lintr}` to [CI](https://github.com/r-lib/actions/tree/master/examples#lint-workflow) (similar to how `cargo fmt` is used in `extendr`)

eitsupi:
(This is probably not an appropriate issue but related to lintr)

The fourth line of the following R file generated by extendr is checked by the commented code linter of lintr, so it is recommended to change it.

```r
# Generated by extendr: Do not edit by hand
#
# This file was created with the following call:
#   .Call("wrap__make_helloextendr_wrappers", use_symbols = TRUE, package_name = "helloextendr")
```

Ilia-Kosenkov:
@eitsupi , I see, but we would like to keep this information in the wrapper (as it is emitted by extendr itself). Is there a way we can mark it so lintr does not check it?

eitsupi:
It is described here.
https://lintr.r-lib.org/articles/lintr.html?q=nolint#excluding-lines-of-code

The easiest way is to add `# nolint` at the end on the line.

```r
# Generated by extendr: Do not edit by hand
#
# This file was created with the following call:
#   .Call("wrap__make_helloextendr_wrappers", use_symbols = TRUE, package_name = "helloextendr") # nolint
```

Ilia-Kosenkov:
I guess it's better to exclude the whole file, since it is generated by extendr. I'll take a look a this soon.

Ilia-Kosenkov:
Hey @eitsupi , can you try out `{rextendr}` from this branch https://github.com/Ilia-Kosenkov/rextendr/tree/exclude-wrappers-from-linting? You can do this like so: `remotes::install_github("Ilia-Kosenkov/rextendr", ref = "exclude-wrappers-from-linting")`
Just install it, and run in your project `rextendr::clean() ; rextendr::document()` to force re-compilation and wrapper generation. Let me know if it solves the linting issue.

eitsupi:
[...]

PR #214, opened 2022-11-21
Exclude wrappers from linting

Ilia-Kosenkov:
See discussion in #169


Ilia-Kosenkov:
So this is a kinda hacky but useful (IMO) fix -- when emitting wrappers, we first right out that 'it was generated by extendr' and then wrap all of the actual code with `nolint` since it is auto-generated. 
I thought of updating `extendr` and how it generates wrappers but decided to do it here in `{rextendr}` as a post-processing step.

PR #217, opened 2022-11-28
Do not compile fake gcc code on Windows

### LINKED ISSUE #209: Unable to mock `libgcc_*` lib file if it already exists
https://github.com/extendr/rextendr/actions/runs/3255637727/jobs/5345165032#step:9:220

> C:\rtools42\x86_64-w64-mingw32.static.posix\bin\ar.exe: creating libgcc_eh.a
> cp: cannot create regular file 'libgcc_s.a': File exists
> ERROR: configuration failed for package 'rextendr'

Ilia-Kosenkov:
Another attempt to close #209. For successful compilation, linker requires `libgcc_eh.a`. Instead of compiling an empty `.c` file into a library file, which is a non-atomic operation, we can resort to simply `touch`ing `libgcc_eh.a`, which satisfies the linker's needs. In the end, it gets all of the necessary exports from other `libgcc*.a` libs, but requires `libgcc_eh.a` to just exist on the search path.


inst/templates/Makevars.win
```diff
@@ -15,10 +15,7 @@ $(SHLIB): $(STATLIB)
 $(STATLIB):
 	mkdir -p $(TARGET_DIR)/libgcc_mock
 	cd $(TARGET_DIR)/libgcc_mock && \
-		touch gcc_mock.c && \
-		gcc -c gcc_mock.c -o gcc_mock.o && \
-		ar -r libgcc_eh.a gcc_mock.o && \
-		cp -n libgcc_eh.a libgcc_s.a
+		touch libgcc_eh.a
```
yutannihilation, line 18: I think this two lines can be more simplified into

```sh
touch $(TARGET_DIR)/libgcc_mock/libgcc_eh.a
```

yutannihilation (commented):
Wow, I didn't notice we didn't need to add `libgcc_s`. Thanks for catching!

Ilia-Kosenkov:
The ci shows I did not fix yet another usage of `cp`, will fix it later today

yutannihilation:
Ah, probably that's the problem? Sorry, it might be my mistake. I can't remember why I added the lines also to `configure.ucrt`. Maybe we can delete the whole file.

Ilia-Kosenkov:
Yep, that's the one. I'll test without it as well.

yutannihilation:
Sorry, I just remembered how it works. On `invoke_cargo()`, the compilation is not what's done inside an R package, so it doesn't use `Makevars.win`. That's why we need another logic to inject the fake `libgcc_eh.a`.

https://github.com/extendr/rextendr/pull/218

PR #218, opened 2022-11-29
Just touch libgcc_eh.a instead of actually compiling it

yutannihilation:
My attempt for #209


Ilia-Kosenkov:
Oh, this looks much better than my attempts! A comment/suggestion: the linker links to `lgcc_eh` and to `lgcc_s`, but in the commenting text you write that it needs `libgcc_eh` and `libgcc_a`. I think the second one has a typo, since it should be `libgcc_eh.a` and `libgcc_s.a`.

Ilia-Kosenkov:
I also restarted the CI. I intend to do it a couple of times in hopes we can reproduce the issue if it is still there.

Ilia-Kosenkov (approved):
Ok, after 3 CI runs it is stable.

yutannihilation:
Thanks for catching! Yes, that was a typo. Since this pull request also removes `libgcc_s` as well, I removed it from the sentences.

PR #219, opened 2022-12-17
[PoC] Do not specify the toolchain on Windows with R >= 4.2

yutannihilation:
Related to https://github.com/extendr/extendr/issues/437

The main purpose of this pull request is to see if it works both with the GNU toolchain and with the MSVC toolchain on Windows with R >= 4.2.

Note that, to support R 4.1, we need to specify `TOOLCHAIN` anyway to make sure the GNU toolchain is used on R 4.1. In other words, if we decide to drop R 4.1, the code can be simplified more.

### DIFF
--- a/.github/workflows/test_pkg_gen.yaml
+++ b/.github/workflows/test_pkg_gen.yaml
@@ -20,8 +20,10 @@ jobs:
       fail-fast: false
       matrix:
         config:
-          - {os: windows-latest, r: 'release', rust-version: 'stable-gnu', rtools-version: '42'}
-          - {os: windows-latest, r: 'devel',   rust-version: 'stable-gnu', rtools-version: '42'}
+          - {os: windows-latest, r: 'release', rust-version: 'stable-gnu',  rtools-version: '42'}
+          - {os: windows-latest, r: 'release', rust-version: 'stable-msvc', rtools-version: '42'}
+          - {os: windows-latest, r: 'devel',   rust-version: 'stable-gnu',  rtools-version: '42'}
+          - {os: windows-latest, r: 'devel',   rust-version: 'stable-msvc', rtools-version: '42'}
           # For R < 4.2, the MSVC toolchain is used to support cross-compilation for the 32-bit.
           # TODO: Remove this runner when we drop the support for R < 4.2
           - {os: windows-latest, r: '4.1',     rust-version: 'stable-msvc'}
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -17,7 +17,9 @@ jobs:
       matrix:
         config:
           - {os: windows-latest, r: 'release', rust-version: 'stable-gnu',  rtools-version: '42'}
+          - {os: windows-latest, r: 'release', rust-version: 'stable-msvc', rtools-version: '42'}
           - {os: windows-latest, r: 'devel',   rust-version: 'stable-gnu',  rtools-version: '42'}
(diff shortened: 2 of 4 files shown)


.github/workflows/test_pkg_gen.yaml
```diff
@@ -20,8 +20,10 @@ jobs:
       fail-fast: false
       matrix:
         config:
-          - {os: windows-latest, r: 'release', rust-version: 'stable-gnu', rtools-version: '42'}
-          - {os: windows-latest, r: 'devel',   rust-version: 'stable-gnu', rtools-version: '42'}
+          - {os: windows-latest, r: 'release', rust-version: 'stable-gnu',  rtools-version: '42'}
+          - {os: windows-latest, r: 'release', rust-version: 'stable-msvc', rtools-version: '42'}
+          - {os: windows-latest, r: 'devel',   rust-version: 'stable-gnu',  rtools-version: '42'}
+          - {os: windows-latest, r: 'devel',   rust-version: 'stable-msvc', rtools-version: '42'}
```
Ilia-Kosenkov, line 26: Since you changed toolchain to `stable` in makevars, I think even if you specify `rust-version: 'stable-gnu'`, it still uses `msvc`. And I think it works because `msvc` is available by default.

Ilia-Kosenkov:
Also, we need to check what toolchain is used for dynamic compilation with `rextendr::rust_source()` and other functions,.

.github/workflows/test_pkg_gen.yaml
yutannihilation: No. The default is set here.

https://github.com/extendr/rextendr/blob/f627e943bfb507192b5599d1e331448c80049906/.github/workflows/R-CMD-check.yaml#L42-L43

Ilia-Kosenkov: Wnat I mean is that in `Makevars` you change toolchain to `msvc`. https://github.com/extendr/rextendr/blob/f627e943bfb507192b5599d1e331448c80049906/inst/templates/Makevars.ucrt#L2

So even if you install `gnu` toolchain, when executing `Makevars`, we are using `msvc`. Unless the env var `TOOLCHAIN` is overriden before.

yutannihilation: Ah, sorry. I thought `stable` can mean GNU or MSVC depending on the default toolchain. I'll fix.

Ilia-Kosenkov:
If this works, I guess we should drop `TOOLCHAIN` from Windows completely. We arrived to using default toolchains on all platforms, so no need to override it via an additional parameter. If there ever is a need to use different toolchain, we should address it by using `rustup default what-we-need`

yutannihilation:
Thanks, sounds a good idea.

yutannihilation:
Hmm, it seems something is wrong with r-hub today so the CI fail... I'll retry tomorrow.

yutannihilation:
Yay, I retried and all the CI passed. I think the current CI setting is a bit overkill; we can drop some combinations that are less important (i.e. mainly test with the MSVC toolchain, as it will be our main target, and keep only one case with GNU toolchain). I'll tweak later.

yutannihilation:
This should be ready for review now.

Ilia-Kosenkov (approved):
Ok, LGTM. Let's make sure that integration tests in `extendr` also succeed with this version of `{rextendr}`.

yutannihilation:
Thanks! I already use this in https://github.com/extendr/extendr/pull/439

PR #221, opened 2022-12-18
Rely on updated pkgbuild::needs_compilation()

### LINKED ISSUE #215: Try relying on the updated mechanism of pkgbuild
As this issue is closed, I need to tackle the issue again...

https://github.com/r-lib/pkgbuild/issues/115

yutannihilation:
Fix #215

(Sorry, I mistakenly committed this directly to the main branch... I reverted)


Ilia-Kosenkov:
Should `track_rust_source` file be removed completely? Do we use its functions anywhere else?


Also, I believe it is worth updating the changelog.

yutannihilation:
Oh, it might be. I'll try it.

I'm still figuring out, but I think this doesn't introduce any user-visible changes. What do you think?

Ilia-Kosenkov:
Hm, you are correct. We used pkgbuild before. The only visible change is the minimum required version :)

yutannihilation:
Yeah. One notable progress is that the developers now can just press the "Load all" menu of RStudio as long as no signatures are changed. But this is not our archivement...

yutannihilation:
> Should `track_rust_source` file be removed completely? Do we use its functions anywhere else?

I checked. In short, there's not many functions that can be removed.

`pretty_rel_path()` is used here; `pretty_rel_single_path()` and `on_error_return_default()` are used inside the function.

https://github.com/extendr/rextendr/blob/ae38dfc0ffa3fef294fc97424c7c19580968a787/R/register_extendr.R#L89

`get_library_path()` is used here.

https://github.com/extendr/rextendr/blob/ae38dfc0ffa3fef294fc97424c7c19580968a787/R/register_extendr.R#L68

`find_newer_files_than()` is used here `get_file_info()` is used inside the function.

https://github.com/extendr/rextendr/blob/ae38dfc0ffa3fef294fc97424c7c19580968a787/R/register_extendr.R#L88

Ilia-Kosenkov:
Sorry I missed your comment here. It feels like `get_file_info()` should no longer be used?

yutannihilation:
No, I meant `get_file_info()` is used in `find_newer_files_than()`, which is used in `register_extendr()`.

https://github.com/extendr/rextendr/blob/c64b0dbd1152c395c82e9191d915c794accc9c27/R/track_rust_source.R#L230-L235

Ilia-Kosenkov:
Ok, now I see. Perhaps it can later be simplified since we only want to check if wrappers are newer than dll or not.

yutannihilation:
Thanks for the hint! I replaced it with bare `file.info()`.

yutannihilation:
Yes, I think so. I was wondering if I can reduce more code, but it seems this is all for now.

R/register_extendr.R
```diff
@@ -49,20 +49,9 @@ register_extendr <- function(path = ".", quiet = FALSE, force = FALSE, compile =
...
-    # This relies on [`pkgbuild::needs_compile()`], which
-    # does not know about Rust files modifications.
-    # `force = TRUE` enforces compilation.
-    pkgbuild::compile_dll(
-      path = path,
-      force = TRUE,
-      quiet = quiet
-    )
+  if (!isFALSE(compile)) {
+    # As of version 1.4.0, pkgbuild can detect the changes on Rust code.
```
Ilia-Kosenkov, line 53:
```suggestion
    # As of version 1.4.0, pkgbuild can detect the changes in Rust code.
```

Ilia-Kosenkov (changes requested):
Looks good, thanks!

Ilia-Kosenkov:
Hm, it blew up because of `{purrr}`. Strange
> DLL 'purrr' not found: maybe not installed for this architecture?

Ilia-Kosenkov (approved):

yutannihilation:
Yeah, I have no idea. I hope time will solve...

PR #224, opened 2022-12-29
[Test] Try the dev version of purrr

yutannihilation:
!!DO NOT MERGE THIS!!

This is to check if the CI failures will be fixed with the dev version of purrr package.


yutannihilation:
Package generation fails for some reason, but now it's confirmed the dev version of purrr package is without the problem. I'm closing.

Issue #173, opened 2022-01-11
RUSTFLAGS


andy-thomason:
Close this if this is already done.

I would suggest `RUSTFLAGS="-C opt-level=3 -C target-cpu=native"` as a default
and a parameter to set them.

These flags make it more likely to get good code performance, especially with loop
vectorisation. We are not building for cross-compilation or a binary package, so
`-C target-cpu=native` will make the best of the local hardware.

We can also create a `.cargo/config.toml` in the project with various options.

It might be worth doing a few benchmarks with different settings.

Ilia-Kosenkov:
So what would be the best approach, temporarily setting env var for compilation or writing `config.toml`? Is there a simple benchmark we can utilize to determine if rust flags are read correctly?

andy-thomason:
I would usually use `objdump -d` to dump the binary and visually inspect the code.

Getting it right can make a 10:1 or better difference in performance. It tends to be a bit
hit-or miss as much depends on whether LLVM will vectorise a loop.

On AVX512 server machines, I've observed substantial performance boosts, especially
when combined with multithreading which can give another two orders of magnitude.

Ilia-Kosenkov:
@andy-thomason, can you please clarify once again, are you suggesting adding `RUSTFLAGS` to dynamic compilation (i.e. `rextendr::rust_function(...)`), or to the templates that we produce for R packages (`rextendr::use_extendr()`)?
If it is the former, won't extra optimizations slow down the compilation, which negatively affects user experience (we had a similar discussion about `dev` vs `release`)?

CGMossa:
There are custom profiles now.

I suggest we make a new configuration, called `optim` or similar that will add this to the `Cargo.toml` file:

```toml
[profile.optim]
inherits = "release"
lto = "thin"
opt-level = 3
panic = "abort"
codegen-units = 1
```

And it can be activated through

```
cargo build --profile optim
```

The settings are picked from various tips & tricks I've seen around.

I'd also suggest we add a `.cargo/config` where 

```toml
[build]
rustflags = ["-Ctarget-cpu=native"]
target-dir = "target"
```

andy-thomason:
Cool!

andy-thomason:
I've been working on the new rust SIMD library to generate the sin, cos, exp, log functions. These would be much faster with `-C target-cpu=native` (ZENV3 or AVX512).

PR #226, opened 2023-01-05
added optimisations to Cargo.toml

alexhroom:
Adds some compiler optimisations to Cargo.toml. This should remove the R CMD CHECK NOTE on the compiled code being very large.


CGMossa:
We have generally decided against this. The gains vs. drawbacks are not the same for every use-case. See this issue for more as to how we decided on this: https://github.com/extendr/rextendr/issues/173 #173

alexhroom:
@CGMossa will create an issue on this topic instead (at @Ilia-Kosenkov's suggestion on discord)- it isn't the same issue as #173 especially now CRAN is more relevant than it was previously

PR #231, opened 2023-01-17
Use Rtools43

yutannihilation:
This pull request is a part of https://github.com/extendr/extendr/issues/442.

r-lib/actions now supports Rtools43 and uses it by default for R > 4.2 (c.f. https://github.com/r-lib/actions/commit/740a0e7ce930b6357f16388c0b2acee8337adb01). So, this pull request removes the specification of `rtools-version: '42'` on the R-devel runners.

Also, this pull request stops relying on pkgbuild to detect `RTOOLS4x_HOME` for R >=4.2. pkgbuild doesn't support Rtools43 yet, and all it does is just to check `RTOOLS4x_HOME` ([source code](https://github.com/r-lib/pkgbuild/blob/11cb9e901582246ff62769d6339e6d1d01a4e4c6/R/rtools.R#L35-L44)), so we can simply do the same thing.

### DIFF
--- a/.github/workflows/test_pkg_gen.yaml
+++ b/.github/workflows/test_pkg_gen.yaml
@@ -22,8 +22,7 @@ jobs:
         config:
           - {os: windows-latest, r: 'release', rust-version: 'stable-gnu',  rtools-version: '42'}
           - {os: windows-latest, r: 'release', rust-version: 'stable-msvc', rtools-version: '42'}
-          - {os: windows-latest, r: 'devel',   rust-version: 'stable-gnu',  rtools-version: '42'}
-          - {os: windows-latest, r: 'devel',   rust-version: 'stable-msvc', rtools-version: '42'}
+          - {os: windows-latest, r: 'devel',   rust-version: 'stable-msvc'}
           # For R < 4.2, the MSVC toolchain is used to support cross-compilation for the 32-bit.
           # TODO: Remove this runner when we drop the support for R < 4.2
           - {os: windows-latest, r: '4.1',     rust-version: 'stable-msvc'}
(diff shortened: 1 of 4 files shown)


.github/workflows/test_pkg_gen.yaml
```diff
@@ -22,8 +22,7 @@ jobs:
         config:
           - {os: windows-latest, r: 'release', rust-version: 'stable-gnu',  rtools-version: '42'}
           - {os: windows-latest, r: 'release', rust-version: 'stable-msvc', rtools-version: '42'}
-          - {os: windows-latest, r: 'devel',   rust-version: 'stable-gnu',  rtools-version: '42'}
```
yutannihilation, line 25: It seems I forgot to remove the extra GNU runner on #219. Let me remove this in this pull reuqest.

R/source.R
```diff
@@ -253,21 +253,33 @@ invoke_cargo <- function(toolchain, specific_target, dir, profile,
...
+        ui_throw("rextendr currently supports R 4.x")
+      }
+
+      if (package_version(R.version$minor) >= "3.0") {
+        rtools_version <- "43"  # nolint: object_usage_linter
+      } else {
+        rtools_version <- "42"  # nolint: object_usage_linter
+      }
+
+      rtools_home <- Sys.getenv(glue("RTOOLS{rtools_version}_HOME"),
```
Ilia-Kosenkov, line 268: Should we also use `normalizePath` here and require it to exist (i.e. `mustWork = TRUE`)?

Ilia-Kosenkov:
Since it is a significant change, I suggest you mention this in the changelog.

yutannihilation:
I see, thanks.

Ilia-Kosenkov (approved):

Issue #238, opened 2023-02-18
pretty_rel_path() fails on Windows with R-devel (R 4.3)


yutannihilation:
This is reproducible on my local.

https://github.com/extendr/extendr/actions/runs/4210717913/jobs/7308589655#step:18:123
```
   ══ Failed tests ════════════════════════════════════════════════════════════════
  ── Failure ('test-helper-functions.R:32'): `pretty_rel_path()` works ───────────
  pretty_rel_path(...) (`actual`) not equal to normalizePath(...) (`expected`).
  
  actual vs expected
  - "R/extendr-wrappers.R"
  + "C:/Users/runneradmin/AppData/Local/Temp/RtmpQ9SQyF/working_dir/RtmpcB6a2L/file6f469214d6a/testpkg/R/extendr-wrappers.R"
  
  [ FAIL 1 | WARN 0 | SKIP 1 | PASS 95 ]
```

Ilia-Kosenkov:
I am looking into this as well right now. I wonder what has changed.

yutannihilation:
It seems the difference is that this line doesn't fail on R-devel

https://github.com/extendr/rextendr/blob/d22584825802c928d8dc1294ea9ec85e2e353a64/R/track_rust_source.R#L39

in case of this

https://github.com/extendr/rextendr/blob/d22584825802c928d8dc1294ea9ec85e2e353a64/tests/testthat/test-helper-functions.R#L33-L36

Ilia-Kosenkov:
Strange, I cannot reproduce it locally with `devtools::test()`, nor with `rcmdcheck::rcmdcheck()`

Ilia-Kosenkov:
Ah wait, found one https://github.com/extendr/rextendr/blob/d22584825802c928d8dc1294ea9ec85e2e353a64/tests/testthat/test-helper-functions.R#L22-L28

```R
>  expect_equal(
+     pretty_rel_path(
+       file.path(pkg_root, "R", "extendr-wrappers.R"),
+       file.path(pkg_root, "src", "rust", "src", "lib.rs")
+     ),
+     "R/extendr-wrappers.R"
+   )
Error: pretty_rel_path(...) (`actual`) not equal to "R/extendr-wrappers.R" (`expected`).

`actual`:   "E:/.../rextendr/R/extendr-wrappers.R"
`expected`: "R/extendr-wrappers.R"
>
```

yutannihilation:
Hmm, it's different one, but probably the same cause?

[...]

PR #240, opened 2023-02-20
Hotfix: Disable flaky test on R-devel-windows

Ilia-Kosenkov:
This disables one flaky test that reliably fails on `R-devel`@Windows. The test covers behavior of `pretty_rel_path()` when the origin path is searched outside of project directory. 
Tracked by #238, but does not resolve it.

A temporary solution to unblock CI/integration tests.


CGMossa:
Lgtm

CGMossa (approved):

PR #244, opened 2023-02-23
Re-enable the test disabled by #240

yutannihilation:
The problem was fixed in R-devel, so we can enable the test again (but probably we have to wait a day or two).

https://github.com/wch/r-source/commit/d444c07b7fe765ce65224faa2e85f0819a189e1a

### DIFF
--- a/tests/testthat/test-pretty_rel_path.R
+++ b/tests/testthat/test-pretty_rel_path.R
@@ -34,8 +34,6 @@ test_that("Find relative path starting outside of package directory, return abso
   pkg_root <- local_package("testpkg")
   use_extendr()
 
-  r_version <- as.numeric_version(glue::glue("{R.version$major}.{R.version$minor}"))
-  is_windows <- .Platform$OS.type == "windows"
 
   expect_equal(
     pretty_rel_path(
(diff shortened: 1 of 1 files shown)


yutannihilation:
The test should fail at the moment, so I canceled the CI for now. I'll rerun it tomorrow.

yutannihilation:
While I confirmed the test passes with the latest R-devel available on https://cran.r-project.org/bin/windows/base/rdevel.html, it seems the GitHub Actions uses a several-days-behind version. Let's wait for some more days.

https://github.com/extendr/rextendr/actions/runs/4252063386/jobs/7412064460#step:9:242

tests/testthat/test-pretty_rel_path.R
```diff
@@ -37,11 +37,6 @@ test_that("Find relative path starting outside of package directory, return abso
   r_version <- as.numeric_version(glue::glue("{R.version$major}.{R.version$minor}"))
   is_windows <- .Platform$OS.type == "windows"
```
Ilia-Kosenkov, line 38: This should be deleted as well since these variables are only used in `skip_if()` condition.
```suggestion
```

yutannihilation: Thanks!

yutannihilation:
The CI now passes. It's ready now.

CGMossa (approved):
LGTM

yutannihilation:
Thanks!

PR #229, opened 2023-01-08
Set lintr for linting R codes

eitsupi:
Addresses #169.

A series of fixes resolved all but one error.

```r
> lintr::lint_package()
.......................................
R/sanitize_code.R:25:1: style: [cyclocomp_linter] Functions should have cyclomatic complexity of less than 15, this has 16.
fill_block_comments <- function(lns, fill_with = " ") {
^
```

All errors can be corrected by refactoring the `fill_block_comments` function or allowing this in the rules.

- For objects used only as strings in `glue::glue` like functions, lintr is not detectable, so the `# nolint` comment has been used.
- `object_name_linter` was turned off because functions that mix underbars and dots are not allowed in lintr.
- `commented_code_linter` was turned off because there is quite a lot of code in the comments.

@Ilia-Kosenkov Sorry to I do this even though you are the assignee of #169. Feel free to edit this PR.

PR #248, opened 2023-03-02
[Bugfix 234]: Fix how `rust_source()` handles code passed via `file` parameter

### LINKED ISSUE #234: `rust_source`: `"wrap__make_rextendr_wrappers" not available for .Call() for package "lib<name>"`
When calling `rust_source` to attempt to include a Rust file, the following error is encountered:
```
> rust_source("src/gpd.rs")
ℹ build directory: /tmp/RtmpZFzz5K/file6b5b3e90ea98
   Compiling gpd v0.0.1 (/tmp/RtmpZFzz5K/file6b5b3e90ea98)
    Finished dev [unoptimized + debuginfo] target(s) in 0.21s
Error in .Call(wrapper_function, use_symbols = use_symbols, package_name = package_name,  :
  "wrap__make_rextendr_wrappers" not available for .Call() for package "libgpd"
```

<details>
<summary>Contents of `src/gpd.rs`</summary>

```rust
use extendr_api::prelude::*;
use extendr_api::wrapper::{ List };


#[extendr]
pub fn grid_row_get_idx(_gr: List, _clump: List) -> Vec<u64>
{
    // Initialise output indices
    let idx: Vec<u64> = Vec::new();

    // Return this as-is, for testing
    idx
}


extendr_module! {
    mod test;
    fn grid_row_get_idx;
}
```
</details>

System info:
- Ubuntu 22.04.1
- R 4.2.0
- rustc stable 1.66.1

-----

A workaround is to use the `code` argument, paired with reading the file contents directly (though there is probably a better way to do this):
```r
rust_source(code = paste0(readLines(path), collapse = "\n"))
```

Ilia-Kosenkov:
Fixes #234 

- `rust_source()` correctly consumes source code from its `file` argument
  - Sourcing the same file multiple times produces new libraries (same as with passing identical code to `code` argument)
  - Sourcing files with identical names but different paths is also supported

- Fix regression introduced in #229 `subset_lines_to_fit_in_limit()`, which affected `ui_throw()`
- Fix redirected link in a vignette
- Remove temporary value assigned to `warning.length` which was needed to bypass `{rlang}` issue
- Add rust toolchain name to the display name of the pipeline


Ilia-Kosenkov:
@yutannihilation , any idea why this is failing? Cannot reproduce locally. I suspect something wrong with the source code on Windows.

Ilia-Kosenkov:
I am currently debugging CI, so the latest runs fail on purpose.

CGMossa:
Is the long-path support a Windows feature? Maybe we should have this check:

```
- name: Check LongPathsEnabled
  run: |
           (Get-ItemProperty "HKLM:System\CurrentControlSet\Control\FileSystem").LongPathsEnabled
```

Ilia-Kosenkov:
For some reason, it does not fail on R-devel on Windows, which may suggest that R-devel does things differently.

CGMossa (approved):

CGMossa (approved):
I feel like these are mainly redundant.

Ilia-Kosenkov:
Yes and no. I am fairly confident in the approach I've chosen, but these test guard us against such problems. And I guarantee you there will be another person using `rust_source()` in a weird scenario, so better be safe than sorry :)

yutannihilation:
> R-devel does things differently.

It seems the implementation around file path on Winodws is getting overhauled in R-devel. But, I don't understand the details and how it relates to the difference we observed here... Anyway, good that it passed the CI.

PR #249, opened 2023-03-05
Rework `features` param in `rust_source()`

### LINKED ISSUE #239: `rust_source()`'s `features` argument seems to be misleading
See this https://github.com/extendr/extendr/issues/481#issuecomment-1435708706 issue for reference. 

Currently, any arguments passed to `features` simply produce a new `[feature]` block in `Cargo.toml`. While it may be useful (?) to some extent, we are talking here about dynamic compilation, which is unlikely to be a complex project.

Consumers, however, may treat `features` as `features = "ndarray"` enabling `extendr-api/ndarray`.

Proposed solutions:
- We drop `features` completely and rely on 
  ```extendr_deps = list(`extendr-api` = list(version = "*", features = array("ndarray", 1)))```, which is a mouthful, and even if documented properly, will be misused
- We keep `features` but instead we go into `extendr_deps`, find `extendr-api` and add `features` to an existing list of features (or create a new one). This should be a reliable thing that works even if `extendr_deps` is provided by the user.

Personally I prefer the latter, since in the majority of cases it will just simplify all the code to `rextendr::rust_source(code = code, features = c("ndarray", "graphics"))`. If we want to allow only selected features, we would have to keep `rextendr` in sync with any features introduced in `extendr`, which is a reasonable tradeoff for configuration safety.

Ilia-Kosenkov:
A general improvement of `rust_source()`. 
- Require `extendr_deps` and `dependencies` to be named lists
- Merge (instead of appending) `extendr_deps` into `dependencies` -- if `dependencies` contains `extendr-api` entry, it will be overwritten by that of `extendr_deps`
- Make `features` a feature flag variable, allowing `"ndarray", "num-complex", "serde", "graphics"` features (and their combinations)
- Correctly set `features` property on `extendr-api` no matter how it is defined
- Add explicit references to relevant packages for all features but `graphics`

Fixed #239


tests/data/either_example.rs
```diff
@@ -0,0 +1,11 @@
+use extendr_api::prelude::*;
+
+#[extendr(use_try_from = true)]
+fn matrix_sum(input : ArrayView2<Rfloat>) -> Rfloat {
```
CGMossa, line 4: You have an `either`-example that doesn't use `either`? 
Lol.. I..... sort of understand why. But maybe we should have a patch.crates-io for `extendr`?
The published version wouldn't really contain and use this example anyways.
You can annotate it nicely?

CGMossa: Right, you can add a `[patch]` section ot the `.cargo/config.toml` or add it via... commandline option..
Thus we "just" need to ignore `.cargo/config.toml` file in the `.Rbuildignore`-file.

CGMossa: Okay. This is a little bit of a headache, because do you maintain `rextendr` so it works with `extendr` published, or `extendr` develop? If you choose the former, than newer features will stagnate and maybe we'll have issues. If you choose the latter then we have to decide to publish `libR-sys`, `extendr` and `rextendr` simultanously.

R/generate_toml.R
```diff
@@ -0,0 +1,93 @@
+merge_named_lists <- function(existing, overwriting,
+                              existign_arg = rlang::caller_arg(existing),
```
multimeric, line 2:
```suggestion
                              existing_arg = rlang::caller_arg(existing),
```

R/generate_toml.R
```diff
@@ -0,0 +1,93 @@
...
+                              overwriting_arg = rlang::caller_arg(overwriting)) {
+  existing <- existing %||% list()
+  existing_names <- unique(names(existing) %||% character(0))
+  if (!is.list(existing) || length(existing_names) != length(existing)) {
+    ui_throw("{.arg {existign_arg}} should be a uniquely named list.")
```
multimeric, line 7:
```suggestion
    ui_throw("{.arg {existing_arg}} should be a uniquely named list.")
```

tests/data/either_example.rs
Ilia-Kosenkov: A little bit overwhelming, but I'll try to respond.
1. It is a mistake. I originally tried using `either` but then remembered it has not been released. I changed code but no the name of the file
2. I track only dependencies. If someone hacks into `patch` or whatever, it is up to that person to define features and imports. My job is to make it `rust_function("fn ndarray_fn() {}", features = "ndarray")`
3. I chose to limit features to a subset defined in the released `extendr`. It maybe unwise. I think I will rework it to not be limited to a fixed set of features, but if an unrecognized feature is passed, I will raise a warning. Thanks for noticing!

R/generate_toml.R
```diff
@@ -0,0 +1,93 @@
+merge_named_lists <- function(existing, overwriting,
```
multimeric, line 1: You may have a good reason for this, but I would strongly advise refactoring this to just use a standard function like `modifyList` or `purrr::list_merge`. I don't like the maintenance burden of adding such a new function ourselves.

R/generate_toml.R
```diff
@@ -0,0 +1,93 @@
...
+    return(extendr_deps)
+  }
+
+  extendr_api <- extendr_deps[["extendr-api"]]
+  if (is.null(extendr_api)) {
+    ui_throw("{.arg extendr_deps} should contain a reference to {.code extendr-api} crate.")
+  }
+
+  if (is.character(extendr_api)) {
+  # extendr-api = "*"
```
multimeric, line 33:
```suggestion
```

R/generate_toml.R
```diff
@@ -0,0 +1,93 @@
...
+    extendr_api <- list(version = extendr_api, features = array(features))
+  } else if (is.list(extendr_api)) {
+  # extendr-api = {git = "extendr/extendr", features = ["feature"]}
```
multimeric, line 36:
```suggestion
```

tests/data/either_example.rs
```diff
@@ -0,0 +1,11 @@
...
+    input.iter().sum()
+}
+
+extendr_module! {
+    mod rextendr;
+    fn matrix_sum;
+}
```
multimeric, line 11:
```suggestion
}

```

tests/testthat/test-optional-features.R
```diff
@@ -0,0 +1,82 @@
+test_that("Feature 'ndarray' is enabled when no extra dependencies are specifiedd", {
```
multimeric, line 1:
```suggestion
test_that("Feature 'ndarray' is enabled when no extra dependencies are specified", {
```

NEWS.md
```diff
@@ -6,6 +6,8 @@
 
 ## New features
 
+* `features` parameter of `rust_source()` now correctly enables features in `extendr-api` and references required crates ([#249](https://github.com/extendr/rextendr/pull/249)).
```
multimeric, line 9: Mention all the argument validation changes?

multimeric (changes requested):

R/generate_toml.R
Ilia-Kosenkov: Oh, it is just my ignorance. `purrr::list_modify()` does exactly what I need, thanks for catching this!

CGMossa:
Please rename the example that doesn't contain either, to something without either.

CGMossa (approved):

multimeric (approved):
Looks good!

R/features.R
```diff
@@ -0,0 +1,65 @@
+features_config <- new.env(parent = emptyenv())
```
multimeric, line 1: Would this be cleaner as something like this? I don't have a strong opinion though.
```suggestion
features_config <- rlang::env(
    known_features = tibble::tribble(
        ~Name, ~RequiresPackage, ~Package,
        "ndarray", TRUE, "ndarray",
        "serde", TRUE, "serde",
        "num-complex", TRUE, "num-complex",
        "graphics", FALSE, NA
    )
)
```
```

File: R/features.R [hunk 11]
```diff
@@ -0,0 +1,65 @@
...
+  "num-complex", TRUE, "num-complex",
+  "graphics", FALSE, NA
+)
+
+validate_extendr_features <- function(features, quiet) {
+  features <- features %||% character(0)
+  vctrs::vec_assert(features, character())
+  features <- unique(features)
+
+  unknown_features <- setdiff(features, features_config[["known_features"]][["Name"]])
```
multimeric, line 15:
```suggestion
  unknown_features <- setdiff(features, features_config$known_features$Name)
```

R/features.R
```diff
@@ -0,0 +1,65 @@
...
+  extendr_deps[["extendr-api"]] <- extendr_api
+
+  extendr_deps
+}
+
+add_features_dependencies <- function(dependencies, features) {
+  required_packages <- features_config[["known_features"]] %>%
+    dplyr::filter(
+      vctrs::vec_in(needles = .data$Name, haystack = features),
+      .data$RequiresPackage
```
multimeric, line 57:
```suggestion
      vctrs::vec_in(needles = .data$Name, haystack = features) &
      .data$RequiresPackage
```

R/features.R
```diff
@@ -0,0 +1,65 @@
...
+  "graphics", FALSE, NA
+)
+
+validate_extendr_features <- function(features, quiet) {
+  features <- features %||% character(0)
+  vctrs::vec_assert(features, character())
+  features <- unique(features)
+
+  unknown_features <- setdiff(features, features_config[["known_features"]][["Name"]])
+  unknown_features <- unknown_features[nzchar(unknown_features)]
```
multimeric, line 16: This is very much optional, but I just find it flows a tiny bit better
```suggestion
  unknown_features <- features %>% 
      setdiff(features_config$known_features$Name) %>%
      # Ignore empty features
      magrittr::extract(., nzchar(.))
```

R/features.R
```diff
@@ -0,0 +1,65 @@
...
+    cli::cli_abort("{.arg extendr_deps} contains an invalid reference to {.code extendr-api} crate.")
+  }
+
+  extendr_deps[["extendr-api"]] <- extendr_api
+
+  extendr_deps
+}
+
+add_features_dependencies <- function(dependencies, features) {
+  required_packages <- features_config[["known_features"]] %>%
```
multimeric, line 54:
```suggestion
  required_packages <- features_config$known_features %>%
```

multimeric (approved):
Basically all my suggestions are style things. I have no actual criticisms.

R/features.R
Ilia-Kosenkov: Fixed manually since github ui disallowed me to commit it from here. The `.` usage causes `{lintr}` to panic, and has no real value here, so I replaced last call by a custom `{vctrs}` - based method with a single arg. If pipelines succeed, I will merge this.

multimeric: Huh, that's weird, I don't see the style issue with that

PR #251, opened 2023-03-07
Add `use_dev_extendr` option

### LINKED ISSUE #250: Allow easy access to development version of `extendr` from `rust_source()` methods
Since `extendr` releases are infrequent, there can be a situation when a desired feature is already available in dev version of `extendr`, but not on crates.io. Currently, when dynamically compiling rust code, this can be overcome by hijacking `extendr_deps` or `patch.crates_io`:
```R
rust_function(
  "fn uses_either() -> either::Either<Rint, Rfloat> { either::Either::Left(Rint::default()) }", 
  features = "either",
  extendr_deps = list(git = "https://github.com/extendr/extendr)
)
```

I propose to provide a shortcut in a form of a boolean flag `use_dev_extendr = FALSE`, which selects `extendr_deps` (stored in options).

So the abovementioned code simplifies to just
```R
rust_function(
  "fn uses_either() -> either::Either<Rint, Rfloat> { either::Either::Left(Rint::default()) }", 
  features = "either",
  use_dev_extendr = TRUE
)
```

I believe this is a dramatic simplification of the API, allowing curious people to explore extendr features right from their R terminal, without the need of a complex project setup.

Note that we will not reduce the flexibility & functionality of `rust_source()`, it will just get an additional parameter, which will impact compilation only if `extendr_deps` are `NULL` ( == not specified by the user). In advanced scenarios, the flag is ignored.

Ilia-Kosenkov:
Add new flag `use_dev_extendr` to `rust_source()` function. If `extendr_deps` is `NULL` (i.e. not specified by the user), populates `extendr_deps` from options, using `rextendr.extendr_deps` if `use_dev_extendr` is not `TRUE`, otherwise, uses `rextendr.extendr_dev_deps`. 

In turn, `rextnedr.extendr_dev_deps` by default is set to ```list(`extendr-api` = list(git = "https://github.com/extendr/extendr"))```.

This allows testing short code snippets against dev extendr with just one extra arg, e.g.

```R
rust_function("fn new_awesome_features_test() {}", use_dev_extendr = TRUE)
```
or, while `either` is not available on `crates.io`, 

```R
rust_function(
  "fn uses_either() -> either::Either<Rint, Rfloat> { either::Either::Left(Rint::default()) }",
  use_dev_extendr = TRUE, 
  features = "either",
  dependencies = list(either = "*")
)
```

Closes #250


CGMossa (approved):
LGTM

PR #259, opened 2023-03-28
Do not add crate reference when feature is enabled

Ilia-Kosenkov:
`extendr` re-exports crates when they are behind a feature gate, so there is no real need to automatically add reference for third-party crates.

Also update how warnings are raised, following the example of `extendr_fn_options`, not raising warnings when `use_dev_extendr = TRUE`.

### DIFF
--- a/R/features.R
+++ b/R/features.R
@@ -2,7 +2,7 @@ features_config <- rlang::env(
   known_features = c("ndarray", "serde", "num-complex", "num-complex", "graphics")
 )
 
-validate_extendr_features <- function(features, quiet) {
+validate_extendr_features <- function(features, suppress_warnings) {
   features <- features %||% character(0)
   vctrs::vec_assert(features, character())
   features <- unique(features)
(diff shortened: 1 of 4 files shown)


multimeric:
This looks good, but is it the *re-exporting* that allows you to simplify this, or is it just the fact that e.g. the `ndarray` feature depends on the `ndarray` crate, making dependency calculation logic redundant? Re-exporting would be a Rust thing whereas this is purely a `cargo.toml` thing, right?

multimeric (approved):

Ilia-Kosenkov:
So the reasoning is the following: somewhere during `either` integration with `extendr` we discussed & decided that it is ok to export third-party crates from `extendr` (I think it is encouraged in Rust community). With that in mind, when a feature is enabled, a respective crate (if any) is automatically re-exported through `extendr` namespace. If you enable `either`, you can immediately use `Either`, `Left`, `Right` if you reference `extendr_api::prelude::*`. Thus, there is no need to install crate dependency explicitly, and I removed this feature in this PR. As a result, `{rextendr}` logic is simplified. 

Of course you can reference any crate explicitly, but then you get several 'versions' of it available, one directly referenced, another -- exported through `extendr`. Since we are talking about dynamic compilation here, I seriously doubt there is a scenario when someone enables `ndarray` but then separately references `ndarray` of a different version to achieve something.

multimeric:
Okay, I guess it's indirectly related to re-exporting, because it mean extendr users can use dependencies of features without adding those dependencies as top level dependencies.

PR #260, opened 2023-03-28
[WIP] Use `{covr}`

Ilia-Kosenkov:
TBD

### DIFF
--- a/.github/workflows/test-coverage.yaml
+++ b/.github/workflows/test-coverage.yaml
@@ -1,0 +1,50 @@
+# Workflow derived from https://github.com/r-lib/actions/tree/v2/examples
+# Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
+on:
+  push:
+    branches: [main, master]
+  pull_request:
+    branches: [main, master]
+
+name: test-coverage
+
+jobs:
+  test-coverage:
+    runs-on: ubuntu-latest
+    env:
+      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: r-lib/actions/setup-r@v2
+        with:
+          use-public-rspm: true
+
+      - uses: r-lib/actions/setup-r-dependencies@v2
+        with:
+          extra-packages: any::covr
+          needs: coverage
+
+      - name: Test coverage
+        run: |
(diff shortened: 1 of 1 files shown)


Ilia-Kosenkov:
This is an experiment & I want your feedback.

multimeric:
It looks good, I have two questions though:

a) Can we see the report that has hopefully been generated from this PR?, and
b) Will this ever cause the build to fail?

Ilia-Kosenkov:
Good questions. Just to clarify, I saw this tool used a lot in different projects but never configured it myself for a project I contribute on GitHub. Thus
1) Unlikely, until we merge it and get the first coverage for `main` branch. It needs a source of truth to compare to. Initial setup is always rough
2) I think we can configure it, e.g. desired/minimum level of coverage, wether it is optional or not, or even disable it altogether. 

Since I cannot answer these questions without experimenting, my suggestion is to try it out and see if we can get a little bit more visibility of test coverage in our code base. This, and some better public look because we report testing & have high coverage (which is always cool). I have a couple of PRs planned for the end of this week, so we can check if we get any value out of it. If not, we can disable integration.

**Note:** If I understand correctly, submitting code coverage to the platform to get observability and allowing the platform to provide feedback/checks are two completely different features. The first one is widely used in R repositories all over the tidyverse, the second is not, so we can easily disable it.

multimeric:
Uploading coverage is the weird part to me, but if you think it would provide visibility I'm not against it. Mostly I want this for CI coverage reports. Can you get covr to output its report to stdout so we can see it for builds?

Ilia-Kosenkov:
It already prints it out in the CI, if that is what you ask.
https://github.com/extendr/rextendr/actions/runs/4546712566/jobs/8015979450?pr=260#step:5:37
image

multimeric (approved):

yutannihilation:
> all over the tidyverse

I think most of them disables the comment feature. Honestly, I feel codecov's comments a bit annoying as they overwrite the notification. Do you want the comments?

https://github.com/search?q=org%3Atidyverse+comment+path%3Acodecov.yml&type=code

yutannihilation (approved):
Looks good!

Ilia-Kosenkov:
I will go ahead and merge it bypassing failing ci -- these two are R-devel which fail because of https://github.com/extendr/extendr/issues/524

Issue #177, opened 2022-02-08
Migrate to `{rlang}` 1.0.0


Ilia-Kosenkov:
A lot of breaking changes in `{rlang}` release. For instance, all our error messages broke (see img attached). With recently added `{cli}` features we need to rework all of the `ui_*` (and related functions).
![image](https://user-images.githubusercontent.com/8782986/153074387-1b522ec6-af99-4ad6-84d6-5b7bdfb8f09b.png)

JosiahParry:
There is one hiccup that needs to be addressed before this can be done completely. Namely, the function `ui_throw()` which _throws_ an error adds a class `rextendr_error` to the error. The idea behind this is that it can be caught using `tryCatch` later down the line. 

`ui_throw()` is ubiquitous. Replacing `ui_throw()` with `cli::cli_abort()` is identical except for the class. 

The class is tested for in `expect_rextendr_error`. Removing `ui_throw()` necessitates that `expect_error()` replaces the rextendr specific version. 

Do you feel comfortable deviating / altering the `principles.md`?

PR #263, opened 2023-04-15
Remove ui_*() helper functions

JosiahParry:
https://github.com/extendr/rextendr/issues/177

### DIFF
--- a/R/clean.R
+++ b/R/clean.R
@@ -29,14 +29,11 @@ clean <- function(path = ".") {
   )
 
   if (!file.exists(toml_path)) {
-    ui_throw(
+    cli::cli_abort(c(
       "Unable to clean binaries.",
-      c(
-        bullet_w(
-          "{.file Cargo.toml} not found in {.path {rust_folder}}."
-        )
-      )
-    )
+      "!" = "{.file Cargo.toml} not found in {.path {rust_folder}}."
+    ))
+
   }
 
   cargo_envvars <- get_cargo_envvars()
@@ -69,12 +66,12 @@ clean <- function(path = ".") {
     } else {
       err_msg <- exec_result$stderr
     }
-    ui_throw(
-      "Unable to execute {.code cargo clean}.",
-      bullet_x(paste(err_msg, collapse = "\n")),
-      call = caller_env(),
-      glue_open = "{<{",
-      glue_close = "}>}"
+    cli::cli_abort(
+      c(
+        "Unable to execute {.code cargo clean}.",
+        "x" = paste(err_msg, collapse = "\n")
+      ),
+      call = caller_env()
     )
   }
   pkgbuild::clean_dll(path = root)
--- a/R/features.R
+++ b/R/features.R
@@ -12,10 +12,16 @@ validate_extendr_features <- function(features, suppress_warnings) {
     discard_empty()
 
   if (!isTRUE(suppress_warnings) && length(unknown_features) > 0) {
-    cli::cli_warn(c(
-      "Found unknown {.code extendr} feature{?s}: {.val {unknown_features}}.",
-      inf_dev_extendr_used() # nolint: object_usage_linter
-    ))
+    # alerts are to be short 1 liners
+    # these are called separately
+    cli::cli_warn(
+      c(
+        "Found unknown {.code extendr} feature{?s}: {.val {unknown_features}}.",
+        "i" = "Are you using a development version of {.code extendr}?"
+      )
+    ) # nolint: object_usage_linter
+
+
   }
 
   features
--- a/R/find_exports.R
+++ b/R/find_exports.R
@@ -42,13 +42,19 @@ extract_meta <- function(lns) {
       glue_collapse(lns, sep = "\n  "),
       1, 80
     )
-    ui_throw(
-      "Rust code contains invalid attribute macros.",
-      c(
-        bullet_x("No valid {.code fn} or {.code impl} block found in the \\
-          following sample:"),
-        code_sample
-      )
+
+
+    cli::cli_abort(
+      cli::cli_fmt({
+        cli::cli_text(
+          "Rust code contains invalid attribute macros."
+        )
+        cli::cli_alert_danger(
+          "No valid {.code fn} or {.code impl} block found in the \\
+          following sample:"
+        )
+          cli::cli_code(code_sample)
+        })
     )
   }
   result
--- a/R/find_exports.R
+++ b/R/find_exports.R
@@ -2,18 +2,22 @@ find_exports <- function(clean_lns) {
   ids <- find_extendr_attrs_ids(clean_lns)
   start <- ids
   end <- dplyr::lead(ids, default = length(clean_lns) + 1L) - 1L
-  purrr::map2_dfr(
-    start,
-    end,
-    ~ extract_meta(clean_lns[seq(.x, .y, by = 1L)])
-  ) %>%
-    # Keeps only name, type (fn|impl) and lifetime of impl
-    # if present.
-    dplyr::transmute(
-      .data$name,
-      type = dplyr::if_else(is.na(.data$impl), "fn", "impl"),
-      .data$lifetime
-    )
+
+  # instantiate list to fill
+  res <- rlang::new_list(length(start))
+
+  for (i in seq_along(start)) {
+    res[[i]] <- extract_meta(clean_lns[start[i]:end[i]])
+  }
+
+  # bind results together into tibble
+  res <- do.call(dplyr::bind_rows, res)
+
+  # Keeps only name, type (fn|impl) and lifetime of impl if present.
+  res[["type"]] <- dplyr::coalesce(res[["impl"]], res[["fn"]])
+
+  res[c("name", "type", "lifetime")]
+
 }
 
 # Finds lines which contain #[extendr] (allowing additional spaces)
@@ -56,6 +60,7 @@ extract_meta <- function(lns) {
           cli::cli_code(code_sample)
         })
     )
+
   }
   result
 }
--- a/R/install_libR_bindings.R
+++ b/R/install_libR_bindings.R
@@ -19,12 +19,12 @@
 install_libR_bindings <- function(version = "*", force = FALSE, quiet = FALSE, patch.crates_io = NULL) {
   package_dir <- find.package("rextendr")
   if (file.access(package_dir, 2) != 0L) {
-    ui_throw("Cannot write to package location: {package_dir}")
+    cli::cli_abort("Cannot write to package location: {package_dir}")
   }
 
   bindings_file <- file.path(package_dir, "rust", "libR-sys", "src", "bindings.rs")
   if (!isTRUE(force) && isTRUE(file.exists(bindings_file))) {
-    ui_i("libR bindings are already installed. Rerun with `force = TRUE` to force re-install.")
+    cli::cli_alert_info("libR bindings are already installed. Rerun with `force = TRUE` to force re-install.")
     return(invisible())
   }
 
--- a/R/register_extendr.R
+++ b/R/register_extendr.R
@@ -36,15 +36,14 @@ register_extendr <- function(path = ".", quiet = FALSE, force = FALSE, compile =
 
   pkg_name <- pkg_name(path)
 
-  ui_i("Generating extendr wrapper functions for package: {.pkg {pkg_name}}.")
+  cli::cli_alert_info("Generating extendr wrapper functions for package: {.pkg {pkg_name}}.")
 
   entrypoint_c_file <- rprojroot::find_package_root_file("src", "entrypoint.c", path = path)
   if (!file.exists(entrypoint_c_file)) {
-    ui_throw(
+    cli::cli_abort(c(
       "Unable to register the extendr module.",
-      c(
-        bullet_x("Could not find file {.file src/entrypoint.c }."),
-        bullet_o("Are you sure this package is using extendr Rust code?")
+      "x" = "Could not find file {.file src/entrypoint.c}.",
+      "*" = "Are you sure this package is using extendr Rust code?"
       )
     )
   }
@@ -64,12 +63,14 @@ register_extendr <- function(path = ".", quiet = FALSE, force = FALSE, compile =
     msg <- "{library_path} doesn't exist"
     if (isTRUE(compile)) {
       # If it doesn't exist even after compile, we have no idea what happened
-      ui_throw(msg)
+      cli::cli_abort(msg)
     } else {
       # If compile wasn't invoked, it might succeed with explicit "compile = TRUE"
-      ui_throw(
-        msg,
-        bullet_i("You need to compile first, try {.code register_rextendr(compile = TRUE)}.")
+      cli::cli_abort(
+        c(
+          msg,
+          "i" = "You need to compile first, try {.code register_rextendr(compile = TRUE)}."
+        )
       )
     }
   }
@@ -80,7 +81,7 @@ register_extendr <- function(path = ".", quiet = FALSE, force = FALSE, compile =
   # wrapper file by hand) so the user might need to run with `force = TRUE`.
   if (!isTRUE(force) && file.info(outfile)[["mtime"]] > file.info(library_path)[["mtime"]]) {
     rel_path <- pretty_rel_path(outfile, path) # nolint: object_usage_linter
-    ui_i("{.file {rel_path}} is up-to-date. Skip generating wrapper functions.")
+    cli::cli_alert_info("{.file {rel_path}} is up-to-date. Skip generating wrapper functions.")
 
     return(invisible(character(0L)))
   }
@@ -97,9 +98,8 @@ register_extendr <- function(path = ".", quiet = FALSE, force = FALSE, compile =
       quiet = quiet
     ),
     error = function(e) {
-      ui_throw(
-        "Failed to generate wrapper functions.",
-        bullet_x(e[["message"]])
+      rlang::abort(
+        c("Failed to generate wrapper functions.", x = e[["message"]])
       )
     }
   )
--- a/R/sanitize_code.R
+++ b/R/sanitize_code.R
@@ -80,13 +80,12 @@ fill_block_comments <- function(lns, fill_with = " ") { # nolint: object_usage_l
   n_close <- sum(valid_syms[["type"]] == "close")
   # Fails if number of `/*` and `*/` are different.
   if (n_open != n_close) {
-    ui_throw(
+    cli::cli_abort(c(
       "Malformed comments.",
-      c(
-        bullet_x("Number of start {.code /*} and end {.code */} \\
-               delimiters are not equal."),
-        bullet_i("Found {n_open} occurence{?s} of {.code /*}."),
-        bullet_i("Found {n_close} occurence{?s} of {.code */}.")
+       "x" = "Number of start {.code /*} and end {.code */} \\
+               delimiters are not equal.",
+      "i" = "Found {n_open} occurence{?s} of {.code /*}.",
+      "i" = "Found {n_close} occurence{?s} of {.code */}."
       )
     )
   }
@@ -114,12 +113,11 @@ fill_block_comments <- function(lns, fill_with = " ") { # nolint: object_usage_l
     any(to_replace[["type"]][2L * seq_len(n_valid / 2L) - 1L] != "open") ||
       any(to_replace[["type"]][2L * seq_len(n_valid / 2L)] != "close")
   ) {
-    ui_throw(
+    cli::cli_abort(c(
       "Malformed comments.",
-      c(
-        bullet_x("{.code /*} and {.code */} are not paired correctly."),
-        bullet_i("This error may be caused by a code fragment like \\
-               {.code */ ... /*}.")
+      "x" = "{.code /*} and {.code */} are not paired correctly.",
+      "i" = "This error may be caused by a code fragment like \\
+               {.code */ ... /*}."
       )
     )
   }
--- a/R/setup.R
+++ b/R/setup.R
@@ -1,14 +1,14 @@
 rextendr_setup <- function(path = ".", cur_version = NULL) {
   if (!file.exists(file.path(path, "DESCRIPTION"))) {
-    ui_throw(
-      "{.arg package.dir} ({.path {path}}) does not contain a DESCRIPTION"
+    cli::cli_abort(
+      "{.arg path} ({.path {path}}) does not contain a DESCRIPTION"
     )
   }
 
   is_first <- is.na(rextendr_version(path))
 
   if (is_first) {
-    ui_i("First time using rextendr. Upgrading automatically...")
+    cli::cli_alert_info("First time using rextendr. Upgrading automatically...")
   }
 
   update_rextendr_version(path, cur_version = cur_version)
@@ -21,12 +21,12 @@ update_rextendr_version <- function(path, cur_version = NULL) {
   prev <- rextendr_version(path)
 
   if (!is.na(cur) && !is.na(prev) && package_version(cur) < package_version(prev)) {
-    ui_w(c(
+    cli::cli_alert_warning(c(
       "Installed rextendr is older than the version used with this package\n",
       "You have {.str {cur}} but you need {.str {prev}}"
     ))
   } else if (!identical(cur, prev)) {
-    ui_i("Setting {.var Config/rextendr/version} to {.str {cur}}")
+    cli::cli_alert_info("Setting {.var Config/rextendr/version} to {.str {cur}}")
     desc::desc_set(`Config/rextendr/version` = cur, file = path)
   }
 }
--- a/R/source.R
+++ b/R/source.R
@@ -127,7 +127,7 @@ rust_source <- function(file, code = NULL,
   dir <- get_build_dir(cache_build)
 
   if (!isTRUE(quiet)) {
-    ui_i("build directory: {.file {dir}}")
+    cli::cli_alert_info("build directory: {.file {dir}}")
   }
 
   # copy rust code into src/lib.rs and determine library name
@@ -297,16 +297,16 @@ invoke_cargo <- function(toolchain, specific_target, dir, profile,
         )
       )
     ) {
-      ui_throw(
+      cli::cli_abort(c(
         "Unable to find Rtools that are needed for compilation.",
-        details = bullet_i("Required version is {.emph {pkgbuild::rtools_needed()}}.")
-      )
+        "i" = "Required version is {.emph {pkgbuild::rtools_needed()}}."
+      ))
     }
 
     if (identical(R.version$crt, "ucrt")) {
       # TODO: update this when R 5.0 is released.
       if (!identical(R.version$major, "4")) {
-        ui_throw("rextendr currently supports R 4.x")
+        cli::cli_abort("rextendr currently supports R 4.x")
       }
 
       if (package_version(R.version$minor) >= "3.0") {
@@ -390,7 +390,7 @@ invoke_cargo <- function(toolchain, specific_target, dir, profile,
 #' @param tty_has_colors \[ logical(1) \] Indicates if output
 #' supports ANSI sequences. If `FALSE`, ANSI sequences are stripped off.
 #' @return \[ character(n) \] Vector of strings
-#' that can be passed to `ui_*`, `cli` or `glue` functions.
+#' that can be passed to `cli` or `glue` functions.
 #' @noRd
 gather_cargo_output <- function(json_output, level, tty_has_colors) {
   rendered_output <-
@@ -436,7 +436,7 @@ check_cargo_output <- function(compilation_result, message_buffer, tty_has_color
         "warning",
         tty_has_colors
       ),
-      ui_w
+      cli::cli_alert_warning
     )
   }
 
@@ -450,12 +450,9 @@ check_cargo_output <- function(compilation_result, message_buffer, tty_has_color
       bullet_x
     )
 
-    ui_throw(
-      "Rust code could not be compiled successfully. Aborting.",
-      error_messages,
-      call = call,
-      glue_open = "{<{",
-      glue_close = "}>}"
+    cli::cli_abort(
+      c("Rust code could not be compiled successfully. Aborting.",
+      error_messages)
     )
   }
 }
@@ -492,7 +489,7 @@ get_specific_target_name <- function() {
       return("i686-pc-windows-gnu")
     }
 
-    ui_throw("Unknown Windows architecture")
+    cli::cli_abort("Unknown Windows architecture")
   }
 
   return(NULL)
--- a/R/eval.R
+++ b/R/eval.R
@@ -85,12 +85,12 @@ fn {fn_name}() -> Result<Robj> {{
       rlang::is_null(fn_handle) ||
         rlang::is_null(dll_handle)
     ) {
-      ui_throw(
-        "The Rust code fragment is no longer available for execution.",
-        details = c(
-          bullet_i("Code fragment can only be executed once."),
-          bullet_w("Make sure you are not re-using an outdated fragment.")
-        )
+      cli::cli_abort(
+        c(
+          "The Rust code fragment is no longer available for execution.",
+            "i" = "Code fragment can only be executed once.",
+            "!" = "Make sure you are not re-using an outdated fragment."
+          )
       )
     }
 
--- a/R/knitr_engine.R
+++ b/R/knitr_engine.R
@@ -20,7 +20,7 @@ eng_extendrsrc <- function(options) {
 
 eng_impl <- function(options, extendr_engine) {
   if (!requireNamespace("knitr", quietly = TRUE)) {
-    ui_throw("The {.pkg knitr} package is required to run the extendr chunk engine.")
+    cli::cli_abort("The {.pkg knitr} package is required to run the extendr chunk engine.")
   }
 
   if (!is.null(options$preamble)) {
@@ -45,11 +45,11 @@ eng_impl <- function(options, extendr_engine) {
     opts$env <- knitr::knit_global()
   }
 
-  ui_v("Compiling Rust extendr code chunk...")
+  cli::cli_alert_success("Compiling Rust extendr code chunk...")
   compiled_code <- do.call(extendr_engine, c(list(code = code), opts))
 
   if (isTRUE(options$eval) && rlang::is_function(compiled_code)) {
-    ui_v("Evaluating Rust extendr code chunk...")
+    cli::cli_alert_success("Evaluating Rust extendr code chunk...")
 
     out <- utils::capture.output({
       result <- withVisible(
--- a/tests/testthat/test-make-module-macro.R
+++ b/tests/testthat/test-make-module-macro.R
@@ -60,31 +60,32 @@ impl Counter {
 })
 
 test_that("Macro generation fails on invalid rust code", {
-  expect_rextendr_error(
+  expect_error(
     make_module_macro("#[extendr]\nlet invalid_var = ();"),
     "Rust code contains invalid attribute macros."
   )
 })
 
+
 test_that("Macro generation fails on invalid comments in code", {
-  expect_rextendr_error(
+  expect_error(
     make_module_macro("/*/*/**/"),
     "Malformed comments."
   )
-  expect_rextendr_error(
+  expect_error(
     make_module_macro("/*/*/**/"),
     "delimiters are not equal"
   )
-  expect_rextendr_error(
+  expect_error(
     make_module_macro("/*/*/**/"),
     "Found 3 occurences"
   )
-  expect_rextendr_error(
+  expect_error(
     make_module_macro("/*/*/**/"),
     "Found 1 occurence"
   )
 
-  expect_rextendr_error(
+  expect_error(
     make_module_macro("*/  /*"),
     "This error may be caused by a code fragment like",
   )


R/clean.R
```diff
@@ -29,14 +29,11 @@ clean <- function(path = ".") {
   )
 
   if (!file.exists(toml_path)) {
-    ui_throw(
+    cli::cli_abort(c(
```
eitsupi, line 32: I suppose `cli::` is unnecessary because this package imports cli.

https://github.com/extendr/rextendr/blob/232133b32125d2f82748668f6cabe64cddb0a1b5/DESCRIPTION#L45

JosiahParry: In general, I think it's best to fully qualify functions from the namespace. 

https://r-pkgs.org/dependencies-in-practice.html#sec-dependencies-in-imports-r-code

> Our recommended default is to call external functions using the package::function() syntax:

tests/testthat/test-use_extendr.R
JosiahParry, line 1: Note that on line 28 this test fails for me. I get message below. I do not understand how snaps work or what this test does. 

```
+ "Message"
+ "  i First time using rextendr. Upgrading automatically..."
+ "  i Setting `Config/rextendr/version` to \"0.2.0.9000\""
+ "  v Creating 'src/rust/src'."
```

R/find_exports.R
```diff
@@ -2,18 +2,22 @@ find_exports <- function(clean_lns) {
...
+  }
+
+  # bind results together into tibble
+  res <- do.call(dplyr::bind_rows, res)
+
+  # Keeps only name, type (fn|impl) and lifetime of impl if present.
+  res[["type"]] <- dplyr::coalesce(res[["impl"]], res[["fn"]])
+
+  res[c("name", "type", "lifetime")]
+
```
Ilia-Kosenkov, line 20: Any reason rewriting to a `for` loop rather than a `map` function?

R/toml_serialization.R
```diff
@@ -41,16 +41,15 @@ to_toml <- function(...,
   names <- names2(args)
 
   # We disallow unnamed top-level atomic arguments
-  invalid <- which(map_lgl(args, ~ is_atomic(.x) && !is.null(.x)))
+  invalid <- which(!nzchar(names) & vapply(args, is.atomic, logical(1)))
```
Ilia-Kosenkov, line 44: Again, why ditching `purrr` in favor of base R?

Ilia-Kosenkov (commented):
How does this work with `quiet = TRUE`? Previously, our helpers respected this (and some options) so that if say `rust_function()` is executed with `quiet = TRUE`, no warnings are displayed.

R/clean.R
multimeric: Also having a package in the Imports doesn't automatically load or attach the package. We could `@import cli` using roxygen but we aren't currently doing so, so `::` isn't unnecessary anyway.

tests/testthat/test-use_extendr.R
JosiahParry: @Ilia-Kosenkov Thank you for pointing that out. I'm not sure what the difference is between cli and the helpers that causes this. 

The "first time using rextendr" message is coming the from the function `rextendr_setup()` which does not have a quiet argument and the ui helpers dont appear to have any clear approach for accessing a global value. 

I'll address this.

R/find_exports.R
JosiahParry: A few reasons. Primarily is that when an error is encounterred in `map2_dfr()` the error message that is propagated is from purrr first thus failing the testthat expectations.  Note the "in index: 1." message. 

```
Error in `map2()`:
ℹ In index: 1.
Caused by error in `extract_meta()`:
! Rust code contains invalid attribute macros.
✖ No valid `fn` or `impl` block found in the following sample:
#[extendr]
let invalid_var = ();
```

By using a for loop, we avoid this. 

Secondly, the for loop is faster and, to my eyes, more readable than the anonymous function piped into a transmute statement using a data mask.

R/toml_serialization.R
JosiahParry: This was preventative based on the above experience with `extract_meta()`.

tests/testthat/test-extendr_function_options.R
```diff
@@ -70,3 +70,10 @@ test_that(
     )
   }
 )
+
+
+test_that(
+  "`rust_function()` does not emit any messages when `quiet = TRUE`", {
+    expect_no_message(rust_function(code = "fn func() {}", quiet = TRUE))
+  }
+)
```
JosiahParry, line 79: @Ilia-Kosenkov new test to verify that no messages are output when `quiet = TRUE`

JosiahParry:
Resolved issue where `quiet = TRUE` was not being respected by `use_extendr(). Additionally, added another test to ensure that `rust_function(quiet = TRUE)` is respected as well.

A note on the implementation. Previously, `use_extendr()` relied on a usethis option to determine if messages should be suppressed. I've modified it based on https://github.com/r-lib/cli/issues/434 which uses the cli option `cli.default_handler`. If `quiet = TRUE` this option is set to `function(...) { }` to print no messages per recommendation. Then, `on.exit()` restores the previous value (most likely `NULL`).

R/use_extendr.R
```diff
@@ -23,11 +23,17 @@
...
-  usethis_quiet <- getOption("usethis.quiet")
-  on.exit(options(usethis.quiet = usethis_quiet))
-  options(usethis.quiet = quiet)
+
+  # https://github.com/r-lib/cli/issues/434
+
+  if (quiet) {
+    old_cli_handler <- options("cli.default_handler")
+    options("cli.default_handler" = function(...) { })
+    on.exit(options("cli.default_handler" = old_cli_handler))
```
eitsupi, line 34: I think we can use `withr::local_options` here.

https://github.com/extendr/rextendr/blob/232133b32125d2f82748668f6cabe64cddb0a1b5/R/use_extendr.R#L205

JosiahParry: I think we can! Was unaware of `local_options()`. Will adjust.

JosiahParry: Addressed this. Also modified other instances of usethis.quiet to utilize cli options since `usethis` is not used for messaging--i.e. no `ui_*()` functions are being called.

R/features.R
```diff
@@ -12,10 +12,16 @@ validate_extendr_features <- function(features, suppress_warnings) {
     discard_empty()
 
   if (!isTRUE(suppress_warnings) && length(unknown_features) > 0) {
-    cli::cli_warn(c(
-      "Found unknown {.code extendr} feature{?s}: {.val {unknown_features}}.",
-      inf_dev_extendr_used() # nolint: object_usage_linter
```
JosiahParry, line 17: this function printed the line `"Are you using a development version of {.code extendr}?"` and was used only once. I figured it was appropriate to delete it.

R/register_extendr.R
```diff
@@ -28,23 +28,23 @@
 #' @seealso [rextendr::document()]
 #' @export
 register_extendr <- function(path = ".", quiet = FALSE, force = FALSE, compile = NA) {
-  usethis_quiet <- getOption("usethis.quiet")
-  on.exit(options(usethis.quiet = usethis_quiet))
-  options(usethis.quiet = quiet)
+
+  if (quiet) {
+    withr::local_options(list("cli.default_handler" = function(...) { }))
+  }
```
JosiahParry, line 34: per suggestion of @eitsupi using {withr} to set local options. 

Since usethis ui_ functions are not used in this package and the package is only suggested, I am unsure if the package should be relying on a usethis option to silence output. 

All `usethis.quiet` references have been replaced with cli.

R/register_extendr.R
```diff
@@ -28,23 +28,23 @@
...
   entrypoint_c_file <- rprojroot::find_package_root_file("src", "entrypoint.c", path = path)
   if (!file.exists(entrypoint_c_file)) {
-    ui_throw(
+    cli::cli_abort(c(
       "Unable to register the extendr module.",
-      c(
-        bullet_x("Could not find file {.file src/entrypoint.c }."),
-        bullet_o("Are you sure this package is using extendr Rust code?")
+      "x" = "Could not find file {.file src/entrypoint.c}.",
+      "*" = "Are you sure this package is using extendr Rust code?"
```
JosiahParry, line 47: note that `"*"` creates a colored bullet. If a black bullet is important this can be changed to use a more verbose `cli::cli_fmt()`

R/register_extendr.R
```diff
@@ -97,9 +96,8 @@ register_extendr <- function(path = ".", quiet = FALSE, force = FALSE, compile =
       quiet = quiet
     ),
     error = function(e) {
-      ui_throw(
-        "Failed to generate wrapper functions.",
-        bullet_x(e[["message"]])
+      rlang::abort(
```
JosiahParry, line 99: This needs to be changed to cli::cli_abort9)

R/rextendr_document.R
```diff
@@ -9,13 +9,13 @@
 #' @inheritParams devtools::document
 #' @return No return value, called for side effects.
 #' @export
-document <- function(pkg = ".", quiet = getOption("usethis.quiet", FALSE), roclets = NULL) {
+document <- function(pkg = ".", quiet = FALSE, roclets = NULL) {
```
JosiahParry, line 12: setting default to `FALSE` as to not rely on global usethis option per reasons above.

R/source.R
```diff
@@ -390,7 +395,7 @@ invoke_cargo <- function(toolchain, specific_target, dir, profile,
 #' @param tty_has_colors \[ logical(1) \] Indicates if output
 #' supports ANSI sequences. If `FALSE`, ANSI sequences are stripped off.
 #' @return \[ character(n) \] Vector of strings
-#' that can be passed to `ui_*`, `cli` or `glue` functions.
+#' that can be passed to `cli` or `glue` functions.
```
JosiahParry, line 398: note doc change to reference cli

R/toml_serialization.R
```diff
@@ -19,7 +19,7 @@
 #'     one line of the resulting output.
 #' @examples
 #' # Produces [workspace] with no children
-#' to_toml(workspace = )
+#' to_toml(workspace = NULL)
```
JosiahParry, line 22: Adding explicit `NULL` in the documentation. Using an empty argument just feels.... wrong.

principles.md
```diff
@@ -8,30 +8,24 @@ rextendr uses the cli package to format messages to the user, which are generall
...
-| `ui_w()` | warn the user about something (note: this is still condition of class `message`, not `warning`)               |
-| `ui_o()` | indicate that the user has something to do                                                                    |
-| `ui_x()` | tell the user that something has gone wrong (note: this still is a condition of class `message`, not `error`) |
+| `cli::cli_alert_success()` | communicate that rextendr has done something successfully, such as write a file                               |
+| `cli::cli_alert_info()` | provide extra information to the user                                                                         |
+| `cli::cli_alert_warning()` | warn the user about something (note: this is still condition of class `message`, not `warning`)               |
+| `cli::cli_ul()` | indicate that the user has something to do                                                                    |
+| `cli::cli_alert_danger()` | tell the user that something has gone wrong (note: this still is a condition of class `message`, not `error`) |
```
JosiahParry, line 15: I think this is a fair representation of the changes.

tests/testthat/helper.R
```diff
@@ -1,13 +1,3 @@
-#' Does code throw an rextendr_error?
-#'
-#' `expect_rextendr_error()` expects an error of class `rextendr_error`, as
-#' thrown by `ui_throw()`.
-#'
-#' @param ... arguments passed to [testthat::expect_error()]
-expect_rextendr_error <- function(...) {
-  testthat::expect_error(..., class = "rextendr_error")
-}
-
```
JosiahParry, line 10: since we are no longer using `ui_throw()` we do not inherit the `rextendr_error` for errors.

R/register_extendr.R
multimeric: Agreed on both counts. Although as a thought, is https://withr.r-lib.org/reference/with_sink.html preferable here because it will silence all output/messages instead of just `cli` ones? Or are we silencing CLI specifically?

R/function_options.R
```diff
@@ -49,7 +49,7 @@ convert_function_options <- function(options, suppress_warnings) {
   } else if (!isTRUE(suppress_warnings) && length(unknown_option_names) > 0) {
     cli::cli_warn(c(
       "Found unknown {.code extendr} function option{?s}: {.val {unknown_option_names}}.",
-      inf_dev_extendr_used() # nolint: object_usage_linter
+      "i" = "Are you using a development version of {.code extendr}?"
```
multimeric, line 52: Don't need to quote these names.

R/rextendr_document.R
multimeric: This one isn't a problem imo. It will default to `FALSE` if the option hasn't been set anyway.

R/sanitize_code.R
```diff
@@ -80,13 +80,12 @@ fill_block_comments <- function(lns, fill_with = " ") { # nolint: object_usage_l
...
   if (n_open != n_close) {
-    ui_throw(
+    cli::cli_abort(c(
       "Malformed comments.",
-      c(
-        bullet_x("Number of start {.code /*} and end {.code */} \\
-               delimiters are not equal."),
-        bullet_i("Found {n_open} occurence{?s} of {.code /*}."),
-        bullet_i("Found {n_close} occurence{?s} of {.code */}.")
+       "x" = "Number of start {.code /*} and end {.code */} \\
```
multimeric, line 85:
```suggestion
      "x" = "Number of start {.code /*} and end {.code */} \\
```

R/setup.R
```diff
@@ -1,14 +1,14 @@
 rextendr_setup <- function(path = ".", cur_version = NULL) {
   if (!file.exists(file.path(path, "DESCRIPTION"))) {
-    ui_throw(
-      "{.arg package.dir} ({.path {path}}) does not contain a DESCRIPTION"
+    cli::cli_abort(
+      "{.arg path} ({.path {path}}) does not contain a DESCRIPTION"
```
multimeric, line 4: Good catch.

R/source.R
```diff
@@ -429,33 +434,22 @@ check_cargo_output <- function(compilation_result, message_buffer, tty_has_color
...
-      "Rust code could not be compiled successfully. Aborting.",
-      error_messages,
-      call = call,
-      glue_open = "{<{",
-      glue_close = "}>}"
+    cli::cli_abort(
+      cli::cli_fmt({
+      cli::cli_text("Rust code could not be compiled successfully. Aborting.")
+
+      if (!quiet) purrr::walk(error_messages, cli::cli_alert_danger)
```
multimeric, line 449: This is odd. Do you not want to `purrr::map` here? Or `walk` outside of the abort?

R/source.R
```diff
@@ -429,33 +434,22 @@ check_cargo_output <- function(compilation_result, message_buffer, tty_has_color
...
+
```
multimeric, line 450:
```suggestion
      if (!quiet) purrr::walk(error_messages, cli::cli_alert_danger)
```

R/toml_serialization.R
multimeric: Yes, if we're sure that `NULL` is equivalent (and I think it is here), then I prefer this too.

R/toml_serialization.R
multimeric: I'm ruling in favour of tidyverse here. If nothing else it's more concise.

R/toml_serialization.R
```diff
@@ -84,18 +83,18 @@ make_header <- function(nm, arg) {
...
 make_idx_msg <- function(invalid, args_limit = 5L) {
   idx <- paste0( # nolint: object_usage_linter
-    glue("`{utils::head(invalid, args_limit)}`"),
+    glue::glue("`{utils::head(invalid, args_limit)}`"),
     collapse = ", "
   )
   if (length(invalid) > args_limit) {
-    idx <- glue("{idx}, ... ")
+    idx <- glue::glue("{idx}, ... ")
   }
```
multimeric, line 91: This is confusing code, but it wasn't introduced in this PR so is probably out of scope.

R/toml_serialization.R
```diff
@@ -114,9 +113,10 @@ simplify_row <- function(row) {
 format_toml <- function(x, ..., .top_level = FALSE) UseMethod("format_toml")
 
 format_toml.default <- function(x, ..., .top_level = FALSE) {
-  ui_throw(
+  cli::cli_abort(c(
     get_toml_err_msg(),
-    bullet_x("`{typeof(x)}` cannot be converted to toml.")
+    "x" = "`{class(x)}` cannot be converted to toml."
```
multimeric, line 118: This change seems correct, since we're using S3 to dispatch to TOML? @Ilia-Kosenkov can you confirm?

R/use_extendr.R
```diff
@@ -200,9 +213,11 @@ throw_if_invalid_rust_name <- function(name, call = caller_env()) {
 use_rextendr_template <- function(template,
                                   save_as = template,
                                   data = list(),
-                                  quiet = getOption("usethis.quiet", FALSE)) {
+                                  quiet = FALSE) {
+
+  if (quiet) withr::local_options(list("cli.default_handler" = function(...) { }))
```
multimeric, line 218: You are doing this a lot. Can you please refactor this into a custom `with` function instead?

R/find_exports.R
JosiahParry: If you're fine with purrr, then some test expectations need to be altered so that they expect the purrr message not the extendr one.

R/find_exports.R
```diff
@@ -42,14 +46,21 @@ extract_meta <- function(lns) {
...
-    ui_throw(
-      "Rust code contains invalid attribute macros.",
-      c(
-        bullet_x("No valid {.code fn} or {.code impl} block found in the \\
-          following sample:"),
-        code_sample
-      )
+
+
+    cli::cli_abort(
```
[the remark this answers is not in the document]
JosiahParry: I couldn't figure out a way to maintain the formatting the cargo errors. What I had tried collapsed new lines. I can try and find a simpler way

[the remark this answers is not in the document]
multimeric: It's more that I thought `cli_fmt()` just returned a string that would be identical to what `cli_abort()` would output here.

R/features.R
Ilia-Kosenkov: It was used in two places AFAIK, but you can delete it if it is no longer used.

NEWS.md
```diff
@@ -4,6 +4,10 @@
 
 * Support Rtools43 (#231).
 
+* `rextendr` migrates to the use of `cli` for raising errors and warnings.
```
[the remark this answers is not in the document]
Ilia-Kosenkov: Agree

NEWS.md
```diff
@@ -4,6 +4,10 @@
...
+  - **Note** that errors no longer have a class of `rextendr_error`. Based on a search of GitHub, there is no code that checks for this class in errors and should not affect any users of rextendr.
```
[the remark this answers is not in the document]
Ilia-Kosenkov: Yep, let's see if we can bypass this.

R/clean.R
```diff
@@ -29,14 +29,11 @@ clean <- function(path = ".") {
...
       "Unable to clean binaries.",
-      c(
-        bullet_w(
-          "{.file Cargo.toml} not found in {.path {rust_folder}}."
-        )
-      )
-    )
+      "!" = "{.file Cargo.toml} not found in {.path {rust_folder}}."
```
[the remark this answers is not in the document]
Ilia-Kosenkov: Yes we can, the `...` arg from `cli::cli_abort()` can pass `class` to `rlang::abort()`, I verified this locally.

R/find_exports.R
Ilia-Kosenkov: I get the `purrr` argument, but I am not convinced we need to use a plain `for` loop.

R/find_exports.R
```diff
@@ -42,14 +46,21 @@ extract_meta <- function(lns) {
...
+      cli::cli_fmt({
+        cli::cli_text(
+          "Rust code contains invalid attribute macros."
+        )
+        cli::cli_alert_danger(
+          "No valid {.code fn} or {.code impl} block found in the \\
+          following sample:"
+        )
+          cli::cli_code(code_sample)
+        })
```
Ilia-Kosenkov, line 61: I think the formatting is off here, try running `{styler}` for this.

R/install_libR_bindings.R
```diff
@@ -19,12 +19,12 @@
 install_libR_bindings <- function(version = "*", force = FALSE, quiet = FALSE, patch.crates_io = NULL) {
   package_dir <- find.package("rextendr")
   if (file.access(package_dir, 2) != 0L) {
-    ui_throw("Cannot write to package location: {package_dir}")
+    cli::cli_abort("Cannot write to package location: {package_dir}")
```
Ilia-Kosenkov, line 22: Apply `.path` or `.file` formatting to `package_dir`?

R/install_libR_bindings.R
```diff
@@ -19,12 +19,12 @@
...
   }
 
   bindings_file <- file.path(package_dir, "rust", "libR-sys", "src", "bindings.rs")
   if (!isTRUE(force) && isTRUE(file.exists(bindings_file))) {
-    ui_i("libR bindings are already installed. Rerun with `force = TRUE` to force re-install.")
+    cli::cli_alert_info("libR bindings are already installed. Rerun with `force = TRUE` to force re-install.")
```
Ilia-Kosenkov, line 27: Should this be `cli::cli_inform()`? Also, how about `{.code force = TRUE}`?

R/register_extendr.R
```diff
@@ -28,23 +28,23 @@
...
+  if (quiet) {
+    withr::local_options(list("cli.default_handler" = function(...) { }))
+  }
 
   rextendr_setup(path = path)
 
   pkg_name <- pkg_name(path)
 
-  ui_i("Generating extendr wrapper functions for package: {.pkg {pkg_name}}.")
+  cli::cli_alert_info("Generating extendr wrapper functions for package: {.pkg {pkg_name}}.")
```
Ilia-Kosenkov, line 40: Same, what's the difference from `cli::cli_inform()`? 
We do not really need a 1-to-1 match with the existing behavior, it is ok to improve it.

R/register_extendr.R
Ilia-Kosenkov: Note an issue.

R/source.R
```diff
@@ -429,33 +434,22 @@ check_cargo_output <- function(compilation_result, message_buffer, tty_has_color
...
+      tty_has_colors
     )
 
-    ui_throw(
-      "Rust code could not be compiled successfully. Aborting.",
-      error_messages,
-      call = call,
-      glue_open = "{<{",
-      glue_close = "}>}"
+    cli::cli_abort(
```
Ilia-Kosenkov, line 445: This is the most difficult part. We want to preserve the ANSI formatting we get from the cargo output so that the user actually sees useful rust compiler hints explaining what is wrong. Just double check that `rust_function()` emits nicely formatted errors if you mess up with your rust code.

R/toml_serialization.R
Ilia-Kosenkov: This is sketchy toml serialization logic we invented some time ago :-D

R/toml_serialization.R
Ilia-Kosenkov: Yep, we rely on `S3`. Is it worth outputting whole `class()` (could be a vector)? Also, I suspect it should be some `{.code {class(x)}}`.

R/toml_serialization.R
JosiahParry: Resolved.

R/clean.R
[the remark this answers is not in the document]
JosiahParry: totally missed that! ...in that case....would a simple wrapper be wanted? Or are there cases where `rextendr_error` should be raised that others shouldn't be?

R/features.R
JosiahParry: Ope! You're correct. I can revert but remove the `"i" = "....."` portion.

R/find_exports.R
JosiahParry: Is the decision then to use purrr and change the test exectation?

R/features.R
JosiahParry: Reverted. Put `inf_dev_extendr_used()` into a new `R/utils.R` file. Since `R/ui.R` no longer exists there was no clear home for such a utility function.

R/function_options.R
JosiahParry: I opted for quoting all names passed to `cli`. See [cli_bullets()](https://cli.r-lib.org/reference/cli_bullets.html#details) examples.

I think the reason why the examples use quoting is because if you want to have `•` or `!` they need to be quoted. It can be odd for examples that might use `i = ` and following by `"!" = ` such as here https://github.com/extendr/rextendr/blob/afe61e61061d87bf41456ca91fc0c0ce2b5ff75b/R/eval.R#L88


```r
cli::cli_abort(
        c(
          "The Rust code fragment is no longer available for execution.",
            "i" = "Code fragment can only be executed once.",
            "!" = "Make sure you are not re-using an outdated fragment."
          )
      )
```

If you prefer to have this unquoted, please confirm and I will unquote all instances of `"i" = `

R/register_extendr.R
JosiahParry: Great point. I think with_sink() is appropriate for functions like `document()` that call external funcitons. But since, in theory, all messages are being emitted with `cli` they _should be_ functionally equivalent. I'm fine with either.

R/rextendr_document.R
```diff
@@ -9,13 +9,13 @@
...
   try_save_all(quiet = quiet)
 
   withr::local_envvar(devtools::r_env_vars())
 
   register_extendr(path = pkg, quiet = quiet)
 
   rlang::check_installed("devtools")
-  devtools::document(pkg = pkg, roclets = roclets, quiet = FALSE)
+  devtools::document(pkg = pkg, roclets = roclets, quiet = quiet)
```
JosiahParry, line 20: @multimeric @Ilia-Kosenkov I inferred here that it would be desirable to silence `devtools::document()` when `quiet = TRUE`. Can either of you confirm that this is indeed desired? Otherwise it can be reverted.

R/sanitize_code.R
JosiahParry: Ran everything through style! Should all be spick and span!

R/source.R
JosiahParry: Per @Ilia-Kosenkov's comment above. This is a surprisingly tough problem! `walk` is preferable to `map` here since we are using `cli::cli_alert_danger` for its side effects (printing to the console) rather than collecting the output from it. 

`error_messages` also needs to be iterated over otherwise the errors are collapsed and `\n`s are removed. 

Though with the addition of `withr::local_options("cli.default_handler")` I think the `if (!quiet)` can be removed.

R/toml_serialization.R
JosiahParry:
```r
identical(to_toml(workspace = ), to_toml(workspace = NULL))
#> TRUE
```

R/toml_serialization.R
JosiahParry: Yeah. This was a goofy one. Take the example here:

```r
to_toml(cat = structure("Meow", class = "kitty"))
Error in `map2_chr()` at rextendr/R/toml_serialization.R:55:2:
ℹ In index: 1.
Caused by error in `format_toml()` at rextendr/R/toml_serialization.R:57:4:
! Object cannot be serialized.
✖ `kitty` cannot be converted to toml.
```

Previously this would said character cannot be converted to toml. which would be wrong :O

JosiahParry: Resolved

``` r
rextendr::to_toml(cat = structure("Meow", class = c("kitty", "cat")))
#> Error in `map2_chr()` at rextendr/R/toml_serialization.R:55:2:
#> ℹ In index: 1.
#> Caused by error in `format_toml()` at rextendr/R/toml_serialization.R:57:4:
#> ! Object cannot be serialized.
#> ✖ `kitty` and `cat` cannot be converted to toml.

#> Backtrace:
#>      ▆
#>   1. └─rextendr::to_toml(...)
#>   2.   └─purrr::map2_chr(...) at rextendr/R/toml_serialization.R:55:2
#>   3.     └─purrr:::map2_("character", .x, .y, .f, ..., .progress = .progress)
#>   4.       ├─purrr:::with_indexed_errors(...)
#>   5.       │ └─base::withCallingHandlers(...)
#>   6.       ├─purrr:::call_with_cleanup(...)
#>   7.       └─rextendr (local) .f(.x[[i]], .y[[i]], ...)
#>   8.         ├─rextendr:::format_toml(...) at rextendr/R/toml_serialization.R:57:4
#>   9.         └─rextendr:::format_toml.default(...) at rextendr/R/toml_serialization.R:113:15
#>  10.           └─cli::cli_abort(c(get_toml_err_msg(), x = "{.code {class(x)}} cannot be converted to toml.")) at rextendr/R/toml_serialization.R:116:2
#>  11.             └─rlang::abort(...)
```

<sup>Created on 2023-04-17 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/use_extendr.R
JosiahParry: What would this look like for you? `with_quiet(quiet)` ?
Also, should this use `with_sink()` instead of `cli.default_handler`?

JosiahParry:
Alright. Most comments have been resolved. There are two major things that need to be sorted out: `rextendr_error` and `quiet` argument. 


## Error throwing 

The big one first: this pr removes the class `rextendr_error` since the `ui_throw()` function was no longer being used. I was unaware that we can pass in arguments to `rlang::abort()` via `...` so thanks for pointing that out. 

It appears that the consensus is ensure `rextendr_error` class remains so that needs to be reinstituted (and thus `expect_rextendr_error` expectations which is no problem). 

I have two questions: 

1. Do we want to have a simple wrapper (less fancy than `ui_throw()` that adds this class? Or should it be explicitly added. I'm personally in favor of the latter. Any new PRs would need to validate that either the wrapper error function is used or the class is thrown. Idk if you can do that with lintr?
2. Do all errors in the package get the class or only functions that deal with rust and cargo directly? 

## Quiet

A wrapper function is desired to silence output. Again, two questions:

1. Do we prefer `withr::local_options(list(cli.default_handler = function(...) {}))` or `withr::with_sink()`
2. What is the name of the wrapper function?

R/use_extendr.R
Ilia-Kosenkov: `with_quiet(quiet)` looks nice & can be added to the docs as a go-to solution for all future public API functions

R/source.R
Ilia-Kosenkov: Honestly, I would say it is not :-( The issue here is the absence of leading spaces, which `{cli}` is so eager to trim. Perhaps we can leave the old implementation just for this particular case? Currently it works like this
```
#> Error in `invoke_cargo()`:
#> ! Rust code could not be compiled successfully. Aborting.
#> x error: expected one of `:`, `@`, or `|`, found `)`
#>  --> src\lib.rs:3:32
#>   |
#> 3 | fn hello_ilia(nowaythiswillwork) {}
#>   |                                ^ expected one of `:`, `@`, or `|`
#>   |
#>   = note: anonymous parameters are removed in the 2018 edition (see RFC 1685)
#> help: if this is a `self` type, give it a parameter name
#>   |
#> 3 | fn hello_ilia(self: nowaythiswillwork) {}
#>   |               +++++
#> help: if this is a parameter name, give it a type
#>   |
#> 3 | fn hello_ilia(nowaythiswillwork: TypeName) {}
#>   |                                ++++++++++
#> help: if this is a type, explicitly ignore the parameter name
#>   |
#> 3 | fn hello_ilia(_: nowaythiswillwork) {}
#>   |               ++
#> x error: expected `:`
#>  --> src\lib.rs:3:32
#>   |
#> 3 | fn hello_ilia(nowaythiswillwork) {}
#>   |                                ^
#> x error[E0425]: cannot find function `meta__hello_ilia` in this scope
#>  --> src\lib.rs:6:4
#>   |
#> 6 | fn hello_ilia;
#>   |    ^^^^^^^^^^ not found in this scope
#> x error: aborting due to 3 previous errors
```
Or directly in the console with ANSI formatting preserved 
image

Note how much more useful it is for debugging.

R/rextendr_document.R
Ilia-Kosenkov: Makes sense to me.

Ilia-Kosenkov:
>Error throwing

>  Do we want to have a simple wrapper (less fancy than ui_throw() that adds this class? Or should it be explicitly added. I'm personally in favor of the latter. Any new PRs would need to validate that either the wrapper error function is used or the class is thrown. Idk if you can do that with lintr?

No strong opinion, you can do it your way.

>    Do all errors in the package get the class or only functions that deal with rust and cargo directly?

Yep, everything we throw ourselves should be a `rextendr_error`.

> A wrapper function is desired to silence output. Again, two questions:

>  Do we prefer withr::local_options(list(cli.default_handler = function(...) {})) or withr::with_sink()

Not sure. `with_sink()` nukes everything, `cli.default_handler` specifically targets `cli`, so feels like a more precise tool.

> What is the name of the wrapper function?

Not sure I follow, the wrapper for silencing things? No preference here

R/source.R
JosiahParry: Ohhhh!! I see that now. Yeesh! Okay, I'll give this another look.

Ilia-Kosenkov: Check if `rlang::last_trace()` looks nice. If it is, then it is a win !

R/find_exports.R
```diff
@@ -2,18 +2,11 @@ find_exports <- function(clean_lns) {
...
-    # Keeps only name, type (fn|impl) and lifetime of impl
-    # if present.
-    dplyr::transmute(
-      .data$name,
-      type = dplyr::if_else(is.na(.data$impl), "fn", "impl"),
-      .data$lifetime
-    )
+
+  purrr::map2_dfr(start, end, ~ extract_meta(clean_lns[.x:.y])) %>%
+    dplyr::mutate(type = dplyr::coalesce(impl, fn)) %>%
```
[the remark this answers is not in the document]
JosiahParry: always forget this bit. will do.

R/utils.R
```diff
@@ -4,3 +4,24 @@
...
+#'
+#' hello_rust()
+#' hello_rust(quiet = TRUE)
+#' @keywords internal
+with_quiet <- function(quiet) {
+  if (quiet) withr::local_options(
+    list("cli.default_handler" = function(...) {}),
+    .local_envir = parent.frame(n = 1)
+    )
+}
```
Ilia-Kosenkov: How about we add `env = rlang::caller_env()`, like it is [commonly done](https://rlang.r-lib.org/reference/eval_tidy.html)? I am a little bit concerned with this `parent.frame()` call.

JosiahParry: TIL `caller_env()` is equivalent to `parent.frame()` + 1
``` r
rlang::caller_env
#> function (n = 1) 
#> {
#>     parent.frame(n + 1)
#> }
#> <bytecode: 0x1053d8350>
#> <environment: namespace:rlang>
```

<sup>Created on 2023-04-18 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

tests/testthat/test-use_extendr.R
```diff
@@ -80,7 +81,9 @@ test_that("use_rextendr_template() works when usethis not available", {
   mockr::with_mock(
     # mock that usethis not installed
     is_installed = function(...) FALSE,
-    { use_extendr() },
+    {
+      use_extendr()
+    },
     .env = "rextendr"
```
JosiahParry, line 87: {styler} did this. Not me. I can restructure if desired.

principles.md
```diff
@@ -8,30 +8,33 @@ rextendr uses the cli package to format messages to the user, which are generall
...
+    "*" = "Are you sure this package is using extendr Rust code?"
+  ),
+  class = "rextendr_error"
 )
 ```
 
-`ui_throw()` emits an error of class `rextendr_error`. This makes it easier to know that an error is coming from rextendr. It also allows downstream users to catch and handle rextendr errors more easily with `tryCatch(..., rextender_error = function(re) ...)`
+## Silencing messages 
+
+cli is used to verbosely inform the user. Functions that have particularly verbose output should be able to be optionally silenced with a `quiet` argument—see `?rust_source` for example.
```
JosiahParry, line 38: added `with_quiet()` to principles

R/find_exports.R
```diff
@@ -2,18 +2,10 @@ find_exports <- function(clean_lns) {
...
-    # if present.
-    dplyr::transmute(
-      .data$name,
-      type = dplyr::if_else(is.na(.data$impl), "fn", "impl"),
-      .data$lifetime
-    )
+
+  purrr::map2_dfr(start, end, ~ extract_meta(clean_lns[.x:.y])) %>%
+    dplyr::mutate(type = dplyr::coalesce(.data$impl, .data$fn)) %>%
+    dplyr::select(dplyr::all_of(c("name", "type", "lifetime")))
```
JosiahParry, line 8: `.data$` right here is deprecated. Introduced warnings into testthat suite but silenced checks. This approach solves checks and tests.

JosiahParry (commented):
I think everything has been addressed. 

```
── R CMD check results ──────────────────────────── rextendr 0.2.0.9000 ────
Duration: 1m 32.3s

0 errors ✔ | 0 warnings ✔ | 0 notes ✔

R CMD check succeeded
```

R/features.R
```diff
@@ -4,7 +4,17 @@ features_config <- rlang::env(
 
 validate_extendr_features <- function(features, suppress_warnings) {
   features <- features %||% character(0)
-  vctrs::vec_assert(features, character())
+
+  is_char <- try(vctrs::vec_assert(features, character()), silent = TRUE)
```
Ilia-Kosenkov, line 8: Why not `vctrs::vec_is(features, character())`?

JosiahParry: Because I didn't know about it! Will quickly push the change. Also, why not `inherits(features, "character")`?

R/utils.R
```diff
@@ -0,0 +1,34 @@
...
+#' @keywords internal
+with_quiet <- function(quiet) {
+  if (quiet) {
+    withr::local_options(
+      list("cli.default_handler" = function(...) {
+      }),
+      .local_envir = rlang::caller_env()
+    )
+  }
+}
```
Ilia-Kosenkov, line 34: Wait wait wait, sorry, I misguided you. 
```suggestion
with_quiet <- function(quiet, env = rlang::caller_env()) {
  if (quiet) {
    withr::local_options(
      list("cli.default_handler" = function(...) {
      }),
      .local_envir = env
    )
  }
}
```

This way you can actually control how `use_quite()` is used in complex setups.

R/find_exports.R
```diff
@@ -2,18 +2,10 @@ find_exports <- function(clean_lns) {
...
-    # Keeps only name, type (fn|impl) and lifetime of impl
-    # if present.
-    dplyr::transmute(
-      .data$name,
-      type = dplyr::if_else(is.na(.data$impl), "fn", "impl"),
-      .data$lifetime
-    )
+
+  purrr::map2_dfr(start, end, ~ extract_meta(clean_lns[.x:.y])) %>%
+    dplyr::mutate(type = dplyr::coalesce(.data$impl, .data$fn)) %>%
```
Ilia-Kosenkov, line 7: While we are at it, here is an issue:
``` r
 rextendr::rust_source(code = "fn test() {}")
#> i build directory: 'C:/Users/.../AppData/Local/Temp/RtmpyoVw1l/file55b864127a'
#> Error in `dplyr::mutate()`:
#> i In argument: `type = dplyr::coalesce(.data$impl, .data$fn)`.
#> Caused by error in `.data$impl`:
#> ! Column `impl` not found in `.data`.
#> Backtrace:
#>      x
#>   1. +-rextendr::rust_source(code = "fn test() {}")
#>   2. | \-rextendr::make_module_macro(code, module_name)
#>   3. |   \-rextendr:::find_exports(sanitize_rust_code(lines))
#>   4. |     \-... %>% ...
#>   5. +-dplyr::select(., dplyr::all_of(c("name", "type", "lifetime")))
#>   6. +-dplyr::mutate(., type = dplyr::coalesce(.data$impl, .data$fn))
#>   7. +-dplyr:::mutate.data.frame(...)
#>   8. | \-dplyr:::mutate_cols(.data, dplyr_quosures(...), by)
#>   9. |   +-base::withCallingHandlers(...)
#>  10. |   \-dplyr:::mutate_col(dots[[i]], data, mask, new_columns)
#>  11. |     \-mask$eval_all_mutate(quo)
#>  12. |       \-dplyr (local) eval()
#>  13. +-dplyr::coalesce(.data$impl, .data$fn)
#>  14. | \-rlang::list2(...)
#>  15. +-impl
#>  16. +-rlang:::`$.rlang_data_pronoun`(.data, impl)
#>  17. | \-rlang:::data_pronoun_get(...)
#>  18. \-rlang:::abort_data_pronoun(x, call = y)
#>  19.   \-rlang::abort(msg, "rlang_error_data_pronoun_not_found", call = call)
```

<sup>Created on 2023-04-19 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Just FYI. I will open a separate issue for this.

R/setup.R
```diff
@@ -21,12 +22,12 @@ update_rextendr_version <- function(path, cur_version = NULL) {
   prev <- rextendr_version(path)
 
   if (!is.na(cur) && !is.na(prev) && package_version(cur) < package_version(prev)) {
-    ui_w(c(
+    cli::cli_alert_warning(c(
       "Installed rextendr is older than the version used with this package\n",
```
Ilia-Kosenkov, line 26:
```suggestion
      "Installed rextendr is older than the version used with this package",
```

R/toml_serialization.R
```diff
@@ -41,15 +41,17 @@ to_toml <- function(...,
...
   if (length(invalid) > 0) {
-    ui_throw(
-      get_toml_err_msg(),
+    cli::cli_abort(
       c(
-        make_idx_msg(invalid),
-        bullet_i("All top-level values should be named.")
-      )
+        get_toml_err_msg(),
+        "x" = make_idx_msg(invalid),
```
Ilia-Kosenkov, line 51: This was a dirty hack before `{cli}` adoption. I think we can just use `x = "Unnamed arguments found at position(s): {invalid}."` and remove `make_idx_msg()`

principles.md
```diff
@@ -8,30 +8,33 @@ rextendr uses the cli package to format messages to the user, which are generall
 
 | function | purpose                                                                                                       |
 |----------|---------------------------------------------------------------------------------------------------------------|
-| `ui_v()` | communicate that rextendr has done something successfully, such as write a file                               |
-| `ui_i()` | provide extra information to the user                                                                         |
-| `ui_w()` | warn the user about something (note: this is still condition of class `message`, not `warning`)               |
-| `ui_o()` | indicate that the user has something to do                                                                    |
-| `ui_x()` | tell the user that something has gone wrong (note: this still is a condition of class `message`, not `error`) |
+| `cli::cli_alert_success()` | communicate that rextendr has done something successfully, such as write a file                               |
```
Ilia-Kosenkov, line 11:
```suggestion
| `cli::cli_alert_success()` | communicate that rextendr has done something successfully, such as wrote a file                               |
```

principles.md
```diff
@@ -8,30 +8,33 @@ rextendr uses the cli package to format messages to the user, which are generall
...
-Each `ui_*()` function has a corresponding `bullet_*()` function that formats text to use in the message. 
 
-Because each `ui_*()` and `bullet_*()` function is processed through cli, they support [glue-based interpolation](https://cli.r-lib.org/articles/semantic-cli.html#interpolation) and [inline text formatting](https://cli.r-lib.org/articles/semantic-cli.html#inline-text-formatting).
+
+[`{cli}`](https://cli.r-lib.org/) supports [glue-based interpolation](https://cli.r-lib.org/articles/semantic-cli.html#interpolation) and [inline text formatting](https://cli.r-lib.org/articles/semantic-cli.html#inline-text-formatting). 
 
 ## Throwing and testing errors
 
-Pass all errors via `ui_throw()`. You can also add additional details via the `details` argument. This can be a good place to provide more information with `bullet_*()` functions, e.g.
+Pass all errors via `cli::cli_abort()`. Ensure that the class `rextendr_error` is provided. You can also add additional details by providing a named vector. See `?cli::cli_abort()` and `?cli::cli_bulelts()` for the `message` argument.
```
Ilia-Kosenkov, line 23:
```suggestion
Pass all errors via `cli::cli_abort()`. Ensure that the class `rextendr_error` is specified. You can also add details by providing a named vector. See `?cli::cli_abort()` and `?cli::cli_bulelts()` for the `message` argument. 
```

principles.md
Ilia-Kosenkov, line 38:
```suggestion
`{cli}` is used to verbosely inform the user. The user should be able to optionally silence functions that have particularly verbose output by setting `quiet` argument to `TRUE`—see `?rust_source` for example. 
```

principles.md
```diff
@@ -8,30 +8,33 @@ rextendr uses the cli package to format messages to the user, which are generall
...
 )
 ```
 
-`ui_throw()` emits an error of class `rextendr_error`. This makes it easier to know that an error is coming from rextendr. It also allows downstream users to catch and handle rextendr errors more easily with `tryCatch(..., rextender_error = function(re) ...)`
+## Silencing messages 
+
+cli is used to verbosely inform the user. Functions that have particularly verbose output should be able to be optionally silenced with a `quiet` argument—see `?rust_source` for example. 
 
-To test for errors from rextendr, use `expect_rextendr_error()`
+Silencing `cli` output should be done with the helper `with_quiet(quiet)`. When `quiet = TRUE` all `cli` output will be suppressed.
```
Ilia-Kosenkov, line 40:
```suggestion
Silencing `cli` output should be done with the helper `with_quiet(quiet)`. When `quiet = TRUE` all, `cli` output is suppressed.
```

Ilia-Kosenkov (changes requested):
LGTM, but with a couple of suggestions.

R/find_exports.R
JosiahParry: Good note. `extract_meta()` should always return a df with the required columns even if 0 rows.

JosiahParry: See https://github.com/extendr/rextendr/issues/264

Ilia-Kosenkov (approved):

R/find_exports.R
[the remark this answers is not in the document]
multimeric: Ah yes I see. `{.code ...}` in dot points like `cli_abort` is inline and has no newlines. I would suggest replacing the outermost `cli_abort` with `stop()` because it's doing the exact same thing here I believe.

R/utils.R
```diff
@@ -0,0 +1,34 @@
...
+#' hello_rust <- function(..., quiet = FALSE) {
+#'   with_quiet(quiet)
+#'   cli::cli_alert_info("This should be silenced when {.code quiet = TRUE}")
+#' }
+#'
+#' hello_rust()
+#' hello_rust(quiet = TRUE)
+#' }
+#' @keywords internal
+with_quiet <- function(quiet, env = rlang::caller_env()) {
```
multimeric, line 26: Following `withr`'s conventions, this should be called `local_quiet` (I would rather `local_quiet_ci` for clarity). `with_` functions have to take an expression as an argument.

R/source.R
```diff
@@ -429,33 +434,30 @@ check_cargo_output <- function(compilation_result, message_buffer, tty_has_color
...
+
+    rlang::abort(
+      c(
+        "Rust code could not be compiled successfully. Aborting.",
+        unlist(errors_formatted)
+      ),
       call = call,
-      glue_open = "{<{",
-      glue_close = "}>}"
+      class = "rextendr_error"
```
multimeric, line 460: Could this whole expression be one pipeline? The temporary vars don't seem necessary.

R/source.R
```diff
@@ -429,33 +434,30 @@ check_cargo_output <- function(compilation_result, message_buffer, tty_has_color
...
 
-    ui_throw(
-      "Rust code could not be compiled successfully. Aborting.",
+    errors_formatted <- purrr::map_chr(
       error_messages,
+      cli::format_inline,
+      keep_whitespace = TRUE
+    ) %>%
+      # removing double new lines with single new line
+      stringi::stri_replace_all_fixed("\n\n", "\n") %>%
```
multimeric, line 450: Why `stringi` rather than `stringr`?

R/source.R
```diff
@@ -429,33 +434,30 @@ check_cargo_output <- function(compilation_result, message_buffer, tty_has_color
...
+      # ensures that the leading cli style `x` is there
+      rlang::set_names("x")
+
+    rlang::abort(
+      c(
+        "Rust code could not be compiled successfully. Aborting.",
+        unlist(errors_formatted)
```
multimeric, line 457: Why would `errors_formatted` be a list here?

multimeric (changes requested):

R/find_exports.R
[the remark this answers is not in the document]
JosiahParry: This would have to be `rlang::abort()` instead of `stop()` because stop does not provide you the ability to set the error class

R/source.R
JosiahParry: You're right. It no longer is. This can be removed. Previously was setting the names in the purrr call which required the output to be a list as `_chr` suffix does not honor the use of named vectors.

R/source.R
JosiahParry: I suppose? I'm not sure how I would do it. It might look something like the below. I'm not sure if it is any better?

```r
  purrr::map_chr(
  error_messages,
  cli::format_inline,
  keep_whitespace = TRUE
) %>%
  # removing double new lines with single new line
  stringi::stri_replace_all_fixed("\n\n", "\n") %>%
  # ensures that the leading cli style `x` is there
  rlang::set_names("x") %>%
  { c("Rust code could not be compiled successfully. Aborting.", .)} %>%
  rlang::abort(class = "rextendr_error")
```

JosiahParry: Though I suppose the error message part can be put into a single one along the lines of below which feels reasonable

```r
error_messages <- gather_cargo_output(
      cargo_output,
      "error",
      tty_has_colors
    ) %>% purrr::map_chr(
      cli::format_inline,
      keep_whitespace = TRUE
    ) %>%
      # removing double new lines with single new line
      stringi::stri_replace_all_fixed("\n\n", "\n") %>%
      # ensures that the leading cli style `x` is there
      rlang::set_names("x")

    rlang::abort(
      c(
        "Rust code could not be compiled successfully. Aborting.",
        errors_messages
      ),
      call = call,
      class = "rextendr_error"
    )
```

multimeric: I prefer this. @Ilia-Kosenkov can you make a call please?

R/source.R
multimeric: Good enough reason, thanks.

R/utils.R
JosiahParry: Just pushed a commit changing it to `local_quiet()` now I'm wondering if `local_quiet_cli()` is better. @multimeric @Ilia-Kosenkov ?

R/find_exports.R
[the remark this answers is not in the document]
JosiahParry: Indeed! `(>= 1.0.5)`

R/source.R
Ilia-Kosenkov: +1 to pipeline approach

R/utils.R
Ilia-Kosenkov: `local_quiet_cli()` sounds exactly right to me.

R/source.R
multimeric: (although I note you're missing the final pipe in the above example)

JosiahParry: In which way? It cannot be directly piped to `rlang::abort()` without first adding 
```r
... above pipeline ... %>%
  { c( "Rust code could not be compiled successfully. Aborting.", .) } %>%
  rlang::abort(call = call, class = "rextendr_error")
```

I'm of the mind that if you have to use inline braces in a pipe then it probably ought to be broken. Though this is y'all work not mine! I defer.

JosiahParry: Erm, I suppose I can make it work with `%>% append("Rust code could not be compiled successfully. Aborting", .)`

JosiahParry: stretching my brain!! resolving with `append(..., after = 0)`

multimeric: What's wrong with:
```R
gather_cargo_output(
      cargo_output,
      "error",
      tty_has_colors
    ) %>% purrr::map_chr(
      cli::format_inline,
      keep_whitespace = TRUE
    ) %>%
      # removing double new lines with single new line
      stringi::stri_replace_all_fixed("\n\n", "\n") %>%
      c(
            "Rust code could not be compiled successfully. Aborting.",
            x = errors_messages
       ) %>%
        rlang::abort(
            call = call,
           class = "rextendr_error"
       )
```

Ilia-Kosenkov: Honestly, I would say that this is an overkill. It is ok to build error messages using pipe, but I see not particular usefulness in going this extra mile. Negatively affects readability.

multimeric: Oh, I find this readable but happy for it to stay as two expressions.

JosiahParry: This is where it is at right now 

https://github.com/JosiahParry/rextendr/blob/d3227398750e6eb708744333c3a815c3dd4aab94/R/source.R#L453

JosiahParry:
```r
    gather_cargo_output(
      cargo_output,
      "error",
      tty_has_colors
    ) %>%
      purrr::map_chr(
        cli::format_inline,
        keep_whitespace = TRUE
      ) %>%
      # removing double new lines with single new line
      stringi::stri_replace_all_fixed("\n\n", "\n") %>%
      # ensures that the leading cli style `x` is there
      rlang::set_names("x") %>%
      # prepend the actual error message
      append("Rust code could not be compiled successfully. Aborting.", after = 0) %>%
      rlang::abort(
        call = call,
        class = "rextendr_error"
    )
 ```

Comment (multimeric) [reply R/source.R]: I find `c`  bit clearer because you can set names. But anyway Ilia doesn't like that so maybe just revert to the two expression option.

Comment (JosiahParry) [reply R/source.R]: Resolved.

**Comment by Ilia-Kosenkov** [comment]
I think if you sync with `main` to get latest CIs, it will be OK

File: R/find_exports.R [hunk 50]
```diff
@@ -2,18 +2,10 @@ find_exports <- function(clean_lns) {
...
-  ) %>%
-    # Keeps only name, type (fn|impl) and lifetime of impl
-    # if present.
-    dplyr::transmute(
-      .data$name,
-      type = dplyr::if_else(is.na(.data$impl), "fn", "impl"),
-      .data$lifetime
-    )
+
+  purrr::map2_dfr(start, end, ~ extract_meta(clean_lns[.x:.y])) %>%
```
eitsupi, line 6: This is not a major problem, but it may be worth replacing the `map_df` functions with other functions since they are marked as "Superseded" in purrr 1.0.
https://purrr.tidyverse.org/reference/map_dfr.html

JosiahParry: You know what isn't superseded? for loops! ;) 
I'll change to `purrr::map()` (though this is a great use case for `base::Map()` too! 🫣)

If we want to go with the recommended approach we would use `purrr::list_rbind()` however this would require us to have a hard requirement on `purrr (>= 1.0.0)`  see [release notes](https://github.com/tidyverse/purrr/releases/tag/v1.0.0). Otherwise we'll have to `do.call(rbind, ...)`. How do y'all want to proceed here? 

@Ilia-Kosenkov @multimeric

JosiahParry:
@Ilia-Kosenkov it has been synced with the upstream. Though there's now the question of how to handle the superceding of `purrr::map_dfr()`

JosiahParry:
`_dfr` suffix is used twice in the code base. I think I'd opt for a follow up issue to fix this rather than solving half of it in here.

R/source.R
```diff
@@ -435,29 +435,28 @@ check_cargo_output <- function(compilation_result, message_buffer, tty_has_color
...
+        stringi::stri_replace_all_fixed("\n\n", "\n") %>%
+        # ensures that the leading cli style `x` is there
+        rlang::set_names("x")
+
+      rlang::abort(
+        c(
+          "Rust code could not be compiled successfully. Aborting.",
+          error_messages
+        ),
+        call = call,
```
multimeric, line 458:
```suggestion
        stringi::stri_replace_all_fixed("\n\n", "\n")

      rlang::abort(
        c(
          "Rust code could not be compiled successfully. Aborting.",
          x = error_messages
        ),
        call = call,
```

multimeric (approved):

R/find_exports.R
Ilia-Kosenkov: Not part of this PR, leave it as is. `map_dfr` will not be removed from the package, but only receive critical bugfixes, there is no problem in using it until we resolve it. You can file a separate issue to track this.

R/source.R
JosiahParry: I don't think this is desired. 

See the below. `rlang::abort()` does not preprend the little `x` that are currently there for each error message from cargo. 

image

Compare that with the existing formatting that this code keeps: 
image

And compared with using `cli::cli_abort()`
image

multimeric: Your comment seems to be about something that my suggestion doesn't impact. But still, if this is the case, go for `cli::cli_abort`. I must have misunderstood something.

JosiahParry: Please excuse the screenshots. Reprex does not render these characters properly for me. 
Here is a reprex that you can run that might visualize the differences a bit better.

Essentially, in order to keep cargo formatting and to have the cli style `x` preprended to each error message, each element of the `error_message` vector must have the name `x` hence the `rlang::set_names()` call. Using `cli` will lose all formatting without the messy `cli::cli_fmt()` function that I previously had. Setting `x = error_message` inside of `rlang::abort()` doesn't do anything in this case because `error_messages` is likely > length 1. 

This is the cleanest iteration of cargo error formatting that I could devise. 

```r
first_error <-  "Rust code could not be compiled successfully. Aborting."

# truncated some error messages.
error_messages <- c("\033[1m\033[38;5;9merror\033[39m: expected one of `:`, `@`, or `|`, found `)`\033[22m\n \033[1m\033[38;5;12m--> \033[39m\033[", "\033[1m\033[38;5;9merror\033[39m: expected `:`\033[22m\n \033[1m\033[38;5;12m--> \033[39m\033[22msrc/lib.rs:3:13\n  \033[1m\033[38;5", "\033[1m\033[38;5;9merror[E0425]\033[39m: cannot find function `meta__hello` in this scope\033[22m\n \033[1m\033[38;5;12", "\033[1m\033[38;5;9merror\033[39m: aborting due to 3 previous errors\033[22m\n")

# rlang keeps any formatting that has been provided
# note that each element of `error_messages` gets a bullet
rlang::abort(
  c(first_error, error_messages),
)


# if we set use_cli_format = TRUE and name the second vector val 
# to `x = error_messages` we don't get the nice `x` in front of the
# error as we'd like _and_ we lose formatting provided by cargo
rlang::abort(
  c(first_error, x = error_messages),
  use_cli_format = TRUE
)

# if, however, we name all of the elements `x` and pass it
# directly into `rlang::abort()` we get the formatting we desire
rlang::abort(
  c(first_error, rlang::set_names(error_messages, "x"))
)
```
image

multimeric:
Are you happy for me to merge, @JosiahParry? You probably have better awareness of the status of all the comments than I.

PR #265, opened 2023-04-22
[R-4.3] Fix CI

Ilia-Kosenkov:
See e.g. https://github.com/extendr/libR-sys/pull/163

R-release is now R-4.3, so there are no longer 32-bit versions of R available. 

Simplify setup and configuration, rely on ready-made workflows rather than manual configuration.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -40,7 +40,7 @@ jobs:
         uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ matrix.config.rust-version }}
-          targets: ${{  matrux.config.rust-target }}
+          targets: ${{  matrix.config.rust-target }}
 
       - uses: r-lib/actions/setup-r@v2
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/R-CMD-check.yaml
```diff
@@ -44,10 +36,10 @@ jobs:
 
       - uses: r-lib/actions/setup-pandoc@v2
 
-      - name: Set up Rust
-        uses: dtolnay/rust-toolchain@master
+      - uses: dtolnay/rust-toolchain@master
```
CGMossa, line 39: what difference did this have?

Ilia-Kosenkov: No difference, I just decided that since we have all other steps unnamed, there is no particular reason to keep an exclusive name for rust installation step.

Ilia-Kosenkov:
Note: since I reset cache version, the first CI run will definitely be suuuuper slow

yutannihilation (approved):
Looks good!

Issue #168, opened 2021-12-28
`rust_sitrep()` or something similar


Ilia-Kosenkov:
Help users understand their `cargo` situation.
https://github.com/extendr/rextendr/issues/104

PR #271, opened 2023-05-01
Add a new (experimental) function `write_license_note`

### LINKED ISSUE #236: A function or script to generate `LICENSE.note`
As detailed in this article https://yutani.rbind.io/post/rust-and-cran-repository-policy/#doesnt-describe-the-authorship-and-copyright by @yutannihilation, it may be necessary to generate a B listing the authors and licenses of Rust's dependent packages to comply with the CRAN policy.
Even if we do not consider the CRAN policy, it is sometimes important to be explicit about the source code contained in the binaries if we intend to distribute them somewhere.

So can we include a function or script in the rextendr to generate the LICENSE.note?

I generated a LICENSE.note in prqlr with the following script and it was accepted by CRAN[^1].
https://github.com/eitsupi/prqlr/blob/380b092c100cdf6a9590aceadd388d677cd62502/dev/generate-license-note.R

[^1]: <https://github.com/cran/prqlr/blob/58774043d150baf64048043a27f1b1e9c6dc2e8e/LICENSE.note>

```r
manifest_path <- file.path("src", "rust", "Cargo.toml")
license_note_path <- file.path("LICENSE.note")

note_header <- paste0(
  "The binary compiled from the source code of this package contains the following Rust packages.\n",
  "\n",
  "\n",
  "-------------------------------------------------------------"
)

package_name <- RcppTOML::parseTOML(manifest_path)$package$name

df_license <- processx::run(
  "cargo",
  c("license", "--authors", "--tsv", "--avoid-build-deps", "--manifest-path", manifest_path)
)$stdout |>
  data.table::fread() |>
  dplyr::select(name, repository, authors, license) |>
  dplyr::filter(name != package_name) |>
  dplyr::mutate(
    authors = dplyr::case_when(
      authors == "" ~ paste0(name, " authors"),
      TRUE ~ authors
    ) |>
      stringr::str_remove_all(r"(\ <.+?>)") |>
      stringr::str_replace_all(r"(\|)", ", ")
  )

license_note <- df_license |>
  purrr::pmap_chr(
    \(name, repository, authors, license) {
      paste0(
        "\n",
        "Name:        ", name, "\n",
        "Repository:  ", repository, "\n",
        "Authors:     ", authors, "\n",
        "License:     ", license, "\n",
        "\n",
        "-------------------------------------------------------------"
      )
    }
  )

c(note_header, license_note) |>
  writeLines(license_note_path)
```

I used [cargo-lisence](https://github.com/onur/cargo-license) to make this easier, but I think in @yutannihilation's string2path uses a Rust-independent script.

eitsupi:
Close #236


R/license_note.R
```diff
@@ -0,0 +1,77 @@
...
+      ),
+      class = "rextendr_warning"
+    )
+  }
+
+  note_header <- paste0(
+    "The binary compiled from the source code of this package contains the following Rust packages.\n",
+    "\n",
+    "\n",
+    "-------------------------------------------------------------"
```
JosiahParry, line 35: Could you use `cli::console_width()` instead of a predefined number of `----`? That way users don't run into undesireable wrapping. 

Alternatively `cli::cli_rule()` handles width for you.

eitsupi:
Now that the tests are working properly, I think you can merge this.

R/license_note.R
```diff
@@ -0,0 +1,79 @@
...
+      c(
+        "LICENSE.note already exists.",
+        "If you want to re-generate LICENSE.note, set `force = TRUE` to {.fun write_license_note}."
+      ),
+      class = "rextendr_error"
+    )
+  }
+
+  package_name <- RcppTOML::parseTOML(manifest_file)$package$name
+
```
JosiahParry, line 32: Could there be a processx call or `cargo-license` before this to check to see if it is available?

tests/testthat/helper.R
```diff
@@ -80,3 +80,12 @@ local_proj_set <- function(envir = parent.frame()) {
 cat_file <- function(...) {
   cat(brio::read_file(file.path(...)))
 }
+
+#' Helper function for skip testing.
+#' @param args Character vector, arguments to the `cargo` command. Pass to [processx::run()]'s args param.
+skip_if_cargo_bin <- function(args = "--help") {
+  if (processx::run("cargo", args, error_on_status = FALSE)$status != 0) {
+    message <- paste0("`cargo ", paste0(args, collapse = " "), "` is not available.")
```
JosiahParry, line 88: This would be awesome to have in `write_license_note()`

eitsupi: > This would be awesome to have in `write_license_note()`

I thought of that too, but no such checking mechanism seems to exist in other functions that call `cargo`.

JosiahParry: I suppose that goes back to `rust_sitrep()` https://github.com/extendr/rextendr/issues/168

.github/workflows/R-CMD-check.yaml
```diff
@@ -41,12 +41,15 @@ jobs:
...
+        if: matrix.config.r == 'release'
+        with:
+          crate: cargo-license
+
       - uses: r-lib/actions/setup-r@v2
         with:
           r-version: ${{ matrix.config.r }}
           rtools-version: ${{ matrix.config.rtools-version }}
-          # TODO: enable RSPM when all the packages are available
-          use-public-rspm: false
```
JosiahParry, line 49: `pak` uses RSPM by defualt I believe so this is no longer needed, ya?

R/license_note.R
```diff
@@ -0,0 +1,79 @@
+#' Generate LICENSE.note file.
+#'
+#' LICENSE.note generated by this function lists dependent Rust packages infomation.
+#' To use this function, the [cargo-lincense](https://crates.io/crates/cargo-license) command must be installed.
+#' @param force Logical indicating whether to re-generating LICENSE.note if LICENSE.note is already exists.
+#' @inheritParams register_extendr
+#' @return No return value, called for side effects.
+#' @export
+#' @importFrom dplyr %>%
```
JosiahParry, line 9: I don't think this is needed since it is imported elsewhere 
see https://github.com/extendr/rextendr/blob/9efe3f20e19b9a62c300dd78f04c4dcb6f60a042/R/rextendr.R#L9

tests/testthat/test-license_note.R
```diff
@@ -0,0 +1,13 @@
+test_that("LICENSE.note", {
```
JosiahParry, line 1: Probably a more informative name for the test `"LICENSE.note is generated properly"`

tests/testthat/helper.R
eitsupi: I added a helper function to check cargo sub command.

I don't know if this will be useful for #168.

R/license_note.R
```diff
@@ -0,0 +1,88 @@
...
+  separator <- "-------------------------------------------------------------"
+
+  note_header <- paste0(
+    "The binary compiled from the source code of this package contains the following Rust packages.\n",
+    "\n",
+    "\n",
+    separator
+  )
+
+  package_name <- RcppTOML::parseTOML(manifest_file)$package$name
```
Ilia-Kosenkov, line 68: Especially since you are using it for a very simple one-time operation

R/license_note.R
```diff
@@ -0,0 +1,88 @@
...
+    "cargo",
+    c(
+      "license",
+      "--authors",
+      "--json",
+      "--avoid-build-deps",
+      "--avoid-dev-deps",
+      "--manifest-path", manifest_file
+    )
+  )$stdout %>%
```
Ilia-Kosenkov, line 50: Add some error handling if this fails for some reason?

DESCRIPTION
```diff
@@ -67,6 +67,7 @@ Suggests:
     mockr,
     rmarkdown,
     rstudioapi,
+    RcppTOML,
```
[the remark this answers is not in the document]
eitsupi: RcppTOML has recently been made compliant with TOML 1.0.0, so it is most reasonable to use this, I think.

R/license_note.R
eitsupi: I think it is sufficient of the error indication of processx.

eitsupi:
Ok, I think I can rewrite it in a way that does not rely on the `cargo license` and RcppTOML using the `cargo metadata` command, so I will do that tomorrow.

DESCRIPTION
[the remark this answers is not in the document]
Ilia-Kosenkov: Reliably - no, but we can cover the majority of the cases with regex I think. Just write out that we belive the rust library name is 'this' and we exclude it from license coverage. In the worst-case scenario, we will not exclude the actual library name, and the user will be able to do it manually. I suspect there should be no scenario where we guess the name of the crate wrongly, which coincides with the real dependency crate, which we then wrongly exclude.  So like if we cover 95% of use cases with little to no dependencies, I suspect it is good.

eitsupi:
I have updated the function. It is no longer dependent on `RcppTOML` or `cargo-license`. All it needs is `cargo`.

There was a small difference in the snapshot due to not using `cargo-license`, but I don't think it is a major issue.

eitsupi:
Sorry, I didn't exclude build dependencies. I will take a closer look.

eitsupi:
Replacement by `cargo metadata` failed, it seems that `cargo license` is still necessary.
However, by using `cargo metadata`, `RcppTOML` dependency could be removed.

R/license_note.R
```diff
@@ -68,14 +74,14 @@ write_license_note <- function(path = ".", force = TRUE) {
   package_name <- RcppTOML::parseTOML(manifest_file)$package$name
 
   note_body <- list_license %>%
-    purrr::keep(function(x) x$name != package_name) %>%
+    purrr::keep(function(x) !(x$name %in% package_names)) %>%
```
Ilia-Kosenkov, line 75: Maybe `purrr::discard(~ vctrs::vec_in(.x, package_names))` or something similar?

R/license_note.R
```diff
@@ -37,23 +30,36 @@ write_license_note <- function(path = ".", force = TRUE) {
...
-    jsonlite::parse_json()
+    jsonlite::parse_json() %>%
+    purrr::pluck("packages")
 
   .prep_authors <- function(authors, package) {
-    ifelse(!is.null(authors), authors, paste0(package, " authors")) %>%
-      stringi::stri_replace_all("", regex = r"(\ <.+?>)") %>%
-      stringi::stri_replace_all(", ", regex = r"(\|)")
+    ifelse(!rlang::is_empty(authors), authors, paste0(package, " authors")) %>%
+      stringi::stri_replace_all("", regex = r"(\ <.+?>)")
```
Ilia-Kosenkov: I think the standard parameter order is (string, detected pattern, replacement). There should be an `stri_replace_all_regex()` version to simplify this.

Ilia-Kosenkov:
Will there be any benefit to storing license info in a `tibble` and then mutating it rather than working over disconnected lists?

eitsupi:
Thanks for the review. I updated.

> Will there be any benefit to storing license info in a `tibble` and then mutating it rather than working over disconnected lists?

`cargo-license` can also output data as tsv instead of json, but I think additional packages like `readr` or `data.table` would be needed to read it on the R side.
Here, I parse json and handle lists with the same structure as json, eliminating the need to depend on those packages.

R/license_note.R
```diff
@@ -0,0 +1,96 @@
+#' Generate LICENSE.note file.
+#'
+#' LICENSE.note generated by this function lists dependent Rust packages infomation.
```
Ilia-Kosenkov, line 3: I am not a native English speaker, but `dependent packages` reads to me as packages that depend on the current one, while we are checking packages that the current package depends on. How about 
```suggestion
#' LICENSE.note generated by this function contains information about Rust crate dependencies.
```

R/license_note.R
```diff
@@ -0,0 +1,96 @@
...
+#' To use this function, the [cargo-lincense](https://crates.io/crates/cargo-license) command must be installed.
+#' @param force Logical indicating whether to re-generating LICENSE.note if LICENSE.note is already exists.
```
Ilia-Kosenkov, line 5:
```suggestion
#' @param force Logical indicating whether to re-generate LICENSE.note if LICENSE.note already exists.
```

R/license_note.R
```diff
@@ -0,0 +1,96 @@
...
+#' @inheritParams register_extendr
+#' @return No return value, called for side effects.
+#' @export
+write_license_note <- function(path = ".", quiet = FALSE, force = TRUE) {
+  if (!cargo_command_available(c("license", "--help"))) {
+    cli::cli_abort(
+      c(
+        "The {.code cargo license} command is required to run the {.fun write_license_note} function.",
+        "Please install cargo-license {.url https://crates.io/crates/cargo-license} first."
```
Ilia-Kosenkov, line 14:
```suggestion
        "Please install cargo-license ({.url https://crates.io/crates/cargo-license}) first."
```

R/license_note.R
```diff
@@ -0,0 +1,96 @@
...
+  .prep_authors <- function(authors, package) {
+    ifelse(!is.null(authors), authors, paste0(package, " authors")) %>%
+      stringi::stri_replace_all_regex(r"(\ <.+?>)", "") %>%
+      stringi::stri_replace_all_regex(r"(\|)", ", ")
+  }
+
+  separator <- "-------------------------------------------------------------"
+
+  note_header <- paste0(
+    "The binary compiled from the source code of this package contains the following Rust packages.\n",
```
Ilia-Kosenkov, line 68:
```suggestion
    "The binary compiled from the source code of this package contains the following Rust crates:\n",
```

R/utils.R
```diff
@@ -32,3 +32,17 @@ local_quiet_cli <- function(quiet, env = rlang::caller_env()) {
     )
   }
 }
+
+#' Helper function for check cargo commands.
+#' @param args Character vector, arguments to the `cargo` command. Pass to [processx::run()]'s args param.
```
Ilia-Kosenkov, line 37:
```suggestion
#' @param args Character vector, arguments to the `cargo` command. Passed to [processx::run()]'s args param.
```

R/utils.R
```diff
@@ -32,3 +32,17 @@ local_quiet_cli <- function(quiet, env = rlang::caller_env()) {
...
+#' @return a logical indicating if the command was available.
```
Ilia-Kosenkov, line 38:
```suggestion
#' @return Logical scalar indicating if the command was available.
```

tests/testthat/helper.R
```diff
@@ -80,3 +80,12 @@ local_proj_set <- function(envir = parent.frame()) {
 cat_file <- function(...) {
   cat(brio::read_file(file.path(...)))
 }
+
+#' Helper function for skip testing.
```
Ilia-Kosenkov, line 84:
```suggestion
#' Helper function for skipping tests when cargo subcommand is unavailable
```

Ilia-Kosenkov (changes requested):
LGTM, a couple of wording improvements here and there.

Ilia-Kosenkov (approved):

CGMossa (approved):
LGTM as well.. I of courses waited until @Ilia-Kosenkov did all the work. But atleast i've skimmed it!

R/license_note.R
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  manifest_file <- rprojroot::find_package_root_file("src", "rust", "Cargo.toml", path = path)
+  outfile <- rprojroot::find_package_root_file("LICENSE.note", path = path)
+
+  if (!isTRUE(force) && file.exists(outfile)) {
+    cli::cli_abort(
+      c(
+        "LICENSE.note already exists.",
+        "If you want to re-generate LICENSE.note, set `force = TRUE` to {.fun write_license_note}."
```
JosiahParry, line 27: regenerate is a word that does not need a hyphen between its prefix and root.

JosiahParry (commented):
Just a singular typo. Otherwise looks lovely.

JosiahParry (commented):
Just a singular typo. Otherwise looks lovely.

R/license_note.R
eitsupi: I have no opinion on modifying this, but note that the same wording is in the existing files.

https://github.com/extendr/rextendr/blob/4d079f7638ecd723fa92102f10aba57635df7a5b/.github/workflows/test_pkg_gen.yaml#L121
https://github.com/extendr/rextendr/blob/4d079f7638ecd723fa92102f10aba57635df7a5b/R/register_extendr.R#L18
https://github.com/extendr/rextendr/blob/4d079f7638ecd723fa92102f10aba57635df7a5b/R/rextendr_document.R#L8

Ilia-Kosenkov: This just means we were wrong before :-D Feel free to modify these occurrences or just leave them out to be fixed in some other PR.

JosiahParry: lol ope!

eitsupi: I have replaced all.

PR #272, opened 2023-05-02
refactor: replace map_dfr family to map + bind_rows

### LINKED ISSUE #266: *_dfr() suffixes in purr are superceded
As of purrr v 1.0.0 all `_dfr()` suffixes have been superceded. The recommendation now is to use `purrr::list_rbind()` after using `purrr::map()` though this would require setting a min version to `>= 1.0.0`--but since dplyr is imported I think `bind_rows()` can be used. 

See related https://github.com/extendr/rextendr/pull/263#discussion_r1174589745

This is not an urgent issue but may become a pain point if any issues are further introduced upstream in a later version of purrr.

eitsupi:
Close #266

### DIFF
--- a/R/find_exports.R
+++ b/R/find_exports.R
@@ -8,7 +8,8 @@ find_exports <- function(clean_lns) {
     return(tibble::tibble(name = character(0), type = character(0), lifetime = character(0)))
   }
 
-  purrr::map2_dfr(start, end, ~ extract_meta(clean_lns[.x:.y])) %>%
+  purrr::map2(start, end, ~ extract_meta(clean_lns[.x:.y])) %>%
+    dplyr::bind_rows() %>%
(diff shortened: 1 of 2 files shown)


CGMossa (approved):

eitsupi:
I don't know why you are against setting a minimum version of purrr.

JosiahParry:
Simplifying the dependency tree for users is generally good. Forcing them to upgrade to purrr v1.0—which has breaking changes—may have unintended consequences for the rest of their work. Not everyone has `renv.lock` files for each of their projects and may not know immediately (or have the time to) how to remediate the errors introduced by using purrr 1.x

Ilia-Kosenkov:
I second the suggestion to go `dplyr::bind_rows()`. We already have this dependency, and it is one of the recommended ways of replacing `purrr::map_dfr()`.

eitsupi:
I replaced `list_rbind` to `bind_rows`.

PR #278, opened 2023-05-16
Remove `install_libR_bindings()` function

### LINKED ISSUE #262: `install_libR_bindings()` is unused?
The new test coverage feature revealed that we practically have no tests for `install_libR_bindings()`. Upon further investigation, I noticed that this function was introduced ~3 years ago and has not been touched since then (excluding some occasional cleanups). We do not use it internally, so my question is, is it even used? Does it have any use cases in the current environment? Pinging @multimeric @yutannihilation @CGMossa 
https://github.com/extendr/rextendr/blob/main/R/install_libR_bindings.R

Ilia-Kosenkov:
Closes #262 

This function seems to be unused now. I suggest removing it before the release.
https://github.com/extendr/rextendr/blob/0cfabec1ada370f5511118175b3a1ada2fbc0cf3/R/install_libR_bindings.R#LL19C13-L19C13


CGMossa (approved):
LGTM

Remove it as this shouldn't really be a problem anymore thanks to a lot of efforts into streamlining librsys.

PR #280, opened 2023-05-23
Release prepartation 0.3.0

Ilia-Kosenkov:
TBD

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,15 +1,20 @@
-# rextendr (development version)
+# rextendr 0.3.0
 
 * Ilia Kosenkov is now the official maintainer.
 
-* Support Rtools43 (#231).
+* Josiah Parry is now a contributor.
+
+* Support Rtools43 ([#231](https://github.com/extendr/rextendr/pull/231)).
 
 * `rextendr` has migrated to the use of `cli` for raising errors and warnings.
+
 * Developer note: new helper function `local_quiet_cli()` introduced in `R/utils.R` to simplify silencing cli output.
   
 ## New features
 
-* A new function `write_license_note()` to generate `LICENSE.note` file from `Cargo.toml` (#271).
+* A new function `rust_sitrep()` that prints out a short report on the currently installed Rust toolchain ([#274](https://github.com/extendr/rextendr/pull/274)).
+
+* A new function `write_license_note()` to generate `LICENSE.note` file from `Cargo.toml` ([#271](https://github.com/extendr/rextendr/pull/271)).
 
 * `extendr_fn_options` parameter of `rust_source()` controls what type of options are emitted to `#[extendr()]` attribute ([#252](https://github.com/extendr/rextendr/pull/252)).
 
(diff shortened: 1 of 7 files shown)


CGMossa:
Well, extendr shouldn't. And in fact I'm actually sure that it doesn't **use** it in the sense that it doesn't do anything. It might be in there. It needs to be deleted anyways.

cran-comments.md
```diff
@@ -1,11 +1,25 @@
...
+      Message: Not Found
 
-## R CMD check results
-There were no ERRORs, WARNINGs, or NOTEs.
+This url is valid, but it returns 404 if no Accept header is specified, which is what happens when it is automatically scanned. 
+See https://github.com/rust-lang/crates.io/issues/788 for details.
+
+*  New maintainer:
+      Ilia Kosenkov <ilia.kosenkov@outlook.com>
+    Old maintainer(s):
```
JosiahParry, line 15: indentation here is off (not that it matters too much)

NEWS.md
```diff
@@ -21,13 +26,13 @@ Development configuration is stored as an option named `rextendr.extendr_dev_dep
 
 * A `<pkg_name>-win.def` file containing DLL exports is created by `rextendr::use_extendr()`. It is used during linking phase on Windows and solves the problem of compiling very large projects, such as `polars` ([#212](https://github.com/extendr/rextendr/pull/212))
 
-* Support extendr macro with options (#128).
+* Support extendr macro with options ([#128](https://github.com/extendr/rextendr/pull/128)).
```
eitsupi, line 29: I think it is common practice not to manually link to GitHub issue numbers because `pkgdown` and others do it automatically.

yutannihilation:
> it seems extendr is still using `VECTOR_PTR`,

As @CGMossa answered, it shouldn't because it's already removed by https://github.com/extendr/extendr/pull/512. So, if the test uses the GitHub version of extendr and the GitHub version of libR-sys, it shouldn't have any problem. Does the error mean it uses the release version of extendr with the GitHub version of libR-sys?

Ilia-Kosenkov:
Ha, you are right, we are using dev version of libR-sys but crates version of extendr-api. https://github.com/extendr/rextendr/blob/f8f964ba268f4330188ddeb4255b1db4b676eedb/.github/workflows/R-CMD-check.yaml#L32

I guess I should remove this so we test against crates version.

yutannihilation:
Oh, I see. I think rextendr should test against the crate version because it's probably what the most users want to use. But, it seems there's a reason to use the GitHub version... Can you pin the specific revision before the current `HEAD` for a workaround?

https://github.com/extendr/rextendr/pull/254

This probably means the new version of libR-sys and extendr should be released as soon as possible.

Ilia-Kosenkov:
Ok, how about I test against dev for now, then we release libR-sys & extendr and we switch back to crates version? Here I needed dev version (I think) only to enable various feature tests like support for gated features & extendr attribute arguments.

And yes, we can pin whatever revision we want, since this env var contains basically toml representation of dependency.

JosiahParry:
I don't know the intricacies of the CI system or how extendr really works, but my 2 cents is this: the CI should mimic exactly what is going to be available to the CRAN version of the package. If the CRAN version of the package will not use github dev versions of libR-sys & extender it should utilize the crates version. Otherwise there is a disconnect in expectations from what users will (likely) install

eitsupi:
I recall from tests, I don't think this package is currently tested on CRAN at all.
https://github.com/extendr/rextendr/blob/f8f964ba268f4330188ddeb4255b1db4b676eedb/tests/testthat.R#L8-L10

Ideally, tests that can also be run on CRAN should be modified to run on CRAN.
(But I know this is work that necessarily has to be done at this release.)

Ilia-Kosenkov:
Ready for re-review

NEWS.md
```diff
@@ -1,25 +1,30 @@
-# rextendr (development version)
+# rextendr 0.3.0
```
eitsupi, line 1: I think this is done by `usethis::use_version()`, so no needs to do in this PR.

yutannihilation: Yes. I think it's fine to do it either in this PR or in another PR. I mean, as rextendr's repo doesn't allow you to push the changes to master directly, you'll need to do it in pull request anyway.

Ilia-Kosenkov: Yep, I postponed it since it should happen at a later phase of submission (see #277).

yutannihilation (approved):
Generally looks good to me.

CGMossa (approved):
LGTM

Ilia-Kosenkov:
Ok so given the time & a couple of approves, I will merge this and cautiously proceed with submission.

PR #283, opened 2023-05-31
Switch to dev version after release

Ilia-Kosenkov:
`{rextendr} 0.3.0` is released. Switch to dev version (#277).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,5 @@
+# rextendr (development version)
+
 # rextendr 0.3.0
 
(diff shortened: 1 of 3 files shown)


NEWS.md
```diff
@@ -1,3 +1,5 @@
+# rextendr (development version)
```
CGMossa, line 1: Nice

Ilia-Kosenkov: It is automagically created by `usethis::use_dev_version()` :-D I only called this function and then updated test snapshots :) Automation for the win!

PR #292, opened 2023-06-25
feat!: allow overwriting rextendr templates by `use_extendr`

### LINKED ISSUE #275: Make renaming rextendr package easy
If you want to change the name of an rextendr R package the easiest way to do so is to create a whole new package and copy over the source files. However, it would be niceto have a function that will do that for you. Going through all mentions of the original package name and changing it tough.

eitsupi:
Resolve #275

This PR allows the `use_extendr` function to overwrite existing templates.
If it is not an interactive session, skip without overwriting.


eitsupi:
@JosiahParry Could you take a look at this?

R/use_extendr.R
```diff
@@ -114,19 +120,23 @@ use_extendr <- function(path = ".",
     dependencies = list(`extendr-api` = "*")
   )
 
-  write_file(
-    text = cargo_toml_content,
-    path = file.path("src", "rust", "Cargo.toml"),
-    search_root_from = path,
-    quiet = quiet
-  )
+  if (!file.exists(file.path("src", "rust", "Cargo.toml"))) {
```
Ilia-Kosenkov, line 126: I see some code duplication, can we move this logic into `write_file()` or another helper function?

R/use_extendr.R
```diff
@@ -5,9 +5,10 @@
 #' `"Hello world!"`. It also adds wrapper code so this Rust function can be called from
 #' R with `hello_world()`.
 #'
-#' To avoid possibly messing up your R package, `use_extendr()` will not do anything if
-#' either a directory `src` or a file `R/extendr-wrappers.R` is already present in your
-#' package source.
+#' This function can be called on an existing project;
+#' you will be asked if either the `src` directory or the `R/extendr-wrappers.R` file
+#' are existing. And, even if you select to overwrite them, `Cargo.toml`, `lib.rs`
+#' and `extendr-wrappers.R` will not be overwritten.
```
Ilia-Kosenkov: It is a little bit unclear to me what are we achieving with this? Like, if nothing exists, it will be the same initialization procedure, but if there are rust files, we do not touch them? So if someone modified Makevars, it will be lost? Maybe it is worth mentioning what **will be** modified instead of what will remain untouched?

R/use_extendr.R
```diff
@@ -114,19 +120,23 @@ use_extendr <- function(path = ".",
...
+    )
+  }
 
-  use_rextendr_template(
-    "lib.rs",
-    save_as = file.path("src", "rust", "src", "lib.rs"),
-    quiet = quiet,
-    data = list(mod_name = mod_name)
-  )
+  if (!file.exists(file.path("src", "rust", "src", "lib.rs"))) {
```
Ilia-Kosenkov, line 135: And here

R/use_extendr.R
```diff
@@ -135,12 +145,14 @@ use_extendr <- function(path = ".",
     data = list(mod_name = mod_name)
   )
 
-  use_rextendr_template(
-    "extendr-wrappers.R",
-    save_as = file.path("R", "extendr-wrappers.R"),
-    quiet = quiet,
-    data = list(pkg_name = pkg_name)
-  )
+  if (!file.exists(file.path("R", "extendr-wrappers.R"))) {
```
Ilia-Kosenkov, line 151: And here

tests/testthat/test-use_extendr.R
```diff
@@ -1,4 +1,6 @@
 test_that("use_extendr() sets up extendr files correctly", {
+  skip_if(!requireNamespace("usethis", quietly = TRUE))
```
Ilia-Kosenkov, line 2: Should we just embrace `{usethis}` at this point? We do so much ceremony to check if the package is available, but we rely on its functionality a lot.

Ilia-Kosenkov (commented):
Please also update `NEWS.md` since it is a significant change of the behavior.

R/use_extendr.R
eitsupi: If using a version control system such as Git, I wonder if it is worth listing them all here, since changes can be detected by the version control system.

tests/testthat/test-use_extendr.R
eitsupi: I believe this is a common way to write tests for functions that call optional dependencies.

eitsupi:
> Please also update NEWS.md since it is a significant change of the behavior.

I think I have already done that.

Ilia-Kosenkov:
>I think I have already done that.

Sorry, sometimes I am just blind to the things I see in front of me :)

R/use_extendr.R
eitsupi: Done in 548d94635799d7862b3742daea3b9aef23418f45

R/use_extendr.R
eitsupi: `use_template` will ask the user individually whether or not to overwrite the file, and will not do so without the user's permission.

eitsupi:
The behavior when a file already exists has been completely changed.

If it is not an interactive session, nothing is done with the existing file and no error is raised.
In an interactive session, we will only be prompted if there is an update, and we can decide whether to overwrite the file on a file-by-file basis.

tests/testthat/test-use_extendr.R
eitsupi: Sorry, `skip_if_not_installed` should be used here.

JosiahParry:
I can give this a review tomorrow!

R/use_extendr.R
```diff
@@ -5,9 +5,9 @@
 #' `"Hello world!"`. It also adds wrapper code so this Rust function can be called from
 #' R with `hello_world()`.
 #'
-#' To avoid possibly messing up your R package, `use_extendr()` will not do anything if
-#' either a directory `src` or a file `R/extendr-wrappers.R` is already present in your
-#' package source.
+#' This function can be called on an existing package including rextendr templates;
```
JosiahParry, line 8: I suggest rewtiting to: 

> This function can be called on an existing package using rextendr templates. In this case, you will be asked to overwrite each file.

JosiahParry:
This is the behavior I received with your new branch! I quite like it. At first I hesitated with not having an `overwrite` argument, but it would be really terrible to not realize your `lib.rs` got overwritten as well. 

image

I would, personally, like for there to be a `use_makevars()` function or similar so that those files can be updated. I will make a new issue for it.

eitsupi:
Thanks for reviewing this.

Yeah, overwriting `lib.rs` and `Cargo.toml` is not ideal.
I considered skipping these by default, but I think we should keep the option given the package renaming.

> I would, personally, like for there to be a use_makevars() function or similar so that those files can be updated. I will make a new issue for it.

Creating a separate function may be excessive; how about adding an option to use_extendr so that the behavior can be changed to use only those files?

JosiahParry:
> ...how about adding an option to use_extendr so that the behavior can be changed to use only those files?

My hesitancy here is that once you start adding too many option to a function it becomes a bit like a magical cauldron that needs special incantations to work how you want it--i.e. the purpose becomes less straight forwrd. 

Alternatively, `rextendr:::use_rextendr_template()` could be made part of the public API so that template files are easier to use.

JosiahParry:
Once the typo in the doc is fixed and regenerated in the man page I think its good to go!

JosiahParry (commented):
I "approve" this PR! Looks quite good.

JosiahParry:
I'm not able to actually approve so I'm putting up the bat signal @Ilia-Kosenkov

CGMossa:
I'll review it then; @JosiahParry you should be able to approve.. We'll investigate that too...

eitsupi:
Approval itself can be done by anyone, but the green check is only given to approvals by those who have the permission to merge, I think.

JosiahParry:
Upon further review, why is the `overwrite` argument not provided? Rather it is inferred by `interactive()`. Could we, instead, have `overwrite = NULL` as the default argument so that it asks you by default. Then users can provide the option if they so desire

Ilia-Kosenkov:
If there are still comments, I will merge as soon as they are resolved.

eitsupi:
> Upon further review, why is the `overwrite` argument not provided? Rather it is inferred by `interactive()`. Could we, instead, have `overwrite = NULL` as the default argument so that it asks you by default. Then users can provide the option if they so desire

Indeed, it is annoying to be asked several times if we want to overwrite. I have updated it.

JosiahParry:
@Ilia-Kosenkov all resolved
image

PR #296, opened 2023-06-27
ci(website): enable pkgdown auto dev mode

eitsupi:
See https://pkgdown.r-lib.org/reference/build_site.html#development-mode

It is better to do this setting to avoid confusion between CRAN released and unreleased versions.


Ilia-Kosenkov (approved):
Anything else before I merge?

CGMossa:
Nope. I think you should merge.

PR #298, opened 2023-06-27
feat!: set `SystemRequirements` field automatically

eitsupi:
Currently a message to notice this field (`Please update the system requirement in 'DESCRIPTION' file.`), but I do not think there is any disadvantage to setting it up automatically.

The notation `SystemRequirements: Cargo (rustc package manager)` has been used in `gifski`, `hellorust` etc.
https://github.com/cran/gifski/blob/8bc440a1e3921af1f37a0deb27a01b8274d9b138/DESCRIPTION#L16
https://github.com/cran/hellorust/blob/2265950f01f2bbd87bab26c4f242b9fa8c9817dc/DESCRIPTION#L17

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # rextendr (development version)
 
+* `use_extendr()` and `document()` now set the `SystemRequirements` field of the `DESCRIPTION` file to
+  `Cargo (rustc package manager)` if the field is empty (#298).
+
 # rextend 0.3.1
 
 * Update package templates to work with Rust >= 1.70 (#285)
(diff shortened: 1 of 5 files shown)


CGMossa:
Another great idea. However, this is a "cran" / free field, and could contain
a number of things. Some packages include that you need GNUMake to compile a package.

eitsupi:
> However, this is a "cran" / free field, and could contain
> a number of things. Some packages include that you need GNUMake to compile a package.

Of course, I understand that and therefore will not overwrite it if `SystemRequirements` has already been set. This is as confirmed by the test.

eitsupi:
Is there any reason this can't be merged?

Ilia-Kosenkov (approved):
LGTM, should I merge right now or do you have anything else to modify?

eitsupi:
Thank you for your review. I think you can merge this one right away.

PR #303, opened 2023-07-17
Skip tests with "long path" on Windows with R < 4.3

yutannihilation:
It seems Rust (>= 1.71) has deeper directory structure than before (1.70), and, in some cases, it seems to hit Windows' path limit. Since [long path is supported by Rtools 43 if the OS supports it](https://blog.r-project.org/2023/03/07/path-length-limit-on-windows/), we don't see such failures on R 4.3. This pull request disables these tests in question on R 4.2 on Windows.

https://github.com/extendr/extendr/pull/586#issuecomment-1637936087

### DIFF
--- a/tests/testthat/helper.R
+++ b/tests/testthat/helper.R
@@ -94,3 +94,11 @@ skip_if_cargo_bin <- function(args = "--help") {
     }
   )
 }
+
+#' Helper function for skipping tests when the test possibly fails because of
+#' the path length limit. This only happens on R (<= 4.2) on Windows.
+skip_on_R42_win <- function() {
+  if (.Platform$OS.type == "windows" && getRversion() < "4.3") {
+    skip("Long path is not supported by this version of Rtools.")
+  }
+}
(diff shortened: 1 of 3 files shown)


Ilia-Kosenkov:
Let's define a dedicated method like `skip_on_R42()`

Ilia-Kosenkov (approved):
Thank you!

yutannihilation:
Thanks, defined.

yutannihilation:
Since this function is under `tests/`, I think this linting error is a false-positive, but I added `testthat::` to suppress the error.

```
file=tests/testthat/helper.R,line=102,col=5,[object_usage_linter] no visible global function definition for ‘skip’
```

yutannihilation:
Thanks. I don't fully understand what's happening, but it seems R < 4.2 doesn't support the long path, guessing from this blog post (confession: I only read some part of this, so I might be wrong).

https://blog.r-project.org/2023/03/07/path-length-limit-on-windows/index.html

Issue #304, opened 2023-08-06
use_version()-like function to sync package version between R and rust crate


JosiahParry:
The default package version for an R package using usethis is 0.0.0.9000 and the default package version for the rust source library is 0.1.0. Maintaining versioning between the source and the R package can be easy to forget. It would be great if there was an rextendr function that could modify the Rust crate version along with the R package version. It could use tomleditR and usethis.

JosiahParry:
Thanks! Re: 

> "I think rextendr should not modify other than then extedr-wrappers.R. Cargo.toml is supposed to be edited by the package author, so trying to modify it might cause unexpected results."

When I use `use_version()` from {usethis} I am knowingly invoking the function and doing it manually. As long as this isn't part of some automated process such as `document()`, I don't think it would be invasive.

JosiahParry:
Perhaps, if we do not want to conflict with usethis, the function could be something like `sync_crate_version()` which could take the version in the DESCRIPTION file and use it in the `Cargo.toml`.

Issue #228, opened 2023-01-07
A function to update package version in `src/rust/Cargo.toml`


eitsupi:
At the moment I think the version of Cargo.toml is still 0.1.0 unless it is updated manually, should it be updated in general?
If so, it would be useful to be able to update it along with the R package version with a function like `usethis::use_version`.

_Originally posted by @eitsupi in https://github.com/extendr/helloextendr/pull/16#discussion_r1063984447_

Ilia-Kosenkov:
The question is, is there a scenario when the crate version is different from the package version? 
Since rust code could be only a part of the package, I can imagine someone making changes only to R code and not touching Rust code, which might qualify for bumping R package version but keeping crate version unchanged.

eitsupi:
Perhaps this is not valuable enough, so I will close it.

PR #307, opened 2023-08-09
[DRAFT - feedback needed] Sync crate and R package versions 

JosiahParry:
This PR closes https://github.com/extendr/rextendr/issues/304 and is related to #228.

This draft creates two functions `use_version()` and `sync_version()`. 

- `sync_version()` will set the version of the Rust crate in `src` to match that in the `DESCRIPTION` file. 
- `use_version()` wraps `usethis::use_version()` and 1. calls `usethis::use_version()` to increment the package version and 2. syncs the version with `sync_version()` 

I personally think masking functions from usethis is a little questionable. I like the workflow of: 

- make some significant changes to a package
- usethis::use_version()
- sync_version()

Alternatively, we could change the name of `use_version()` to `increment_version()` to avoid namespace conflicts.


JosiahParry:
CC @eitsupi I would like your eyes on this! I opted for `cargo-bump` since it is easier to use and isn't tied to actually publishing a crate.

eitsupi:
> I opted for `cargo-bump` since it is easier to use and isn't tied to actually publishing a crate.

`cargo-bump` does not seem to be currently maintained and is not recommended to be used.

JosiahParry:
Sure, thats fine. More importantly, though, is the user experience. Do you prefer masking usethis?

eitsupi:
> Do you prefer masking usethis?

I would also prefer to update only Cargo.toml without wrapping the usethis function as you suggested.

PR #308, opened 2023-08-15
feat(template): adds configure scripts

eitsupi:
From <https://cran.r-project.org/web/packages/using_rust.html>

> The configure/configure.win script should check for the presence of commands cargo and rustc, and check their versions if required. This includes checking for system versions on the path and personal versions in ~/.cargo/bin (which are often not on the path). The Linux servers on the CRAN check farm use system versions, and Linux distributions are often slow to update these so version requirements need to be conservative.

configure scripts are copied from prqlr 0.5.0

### DIFF
--- a/R/use_extendr.R
+++ b/R/use_extendr.R
@@ -69,6 +69,22 @@ use_extendr <- function(path = ".",
   }
 
   use_rextendr_template(
+    "configure",
+    save_as = file.path("configure"),
+    quiet = quiet,
+    as_executable = TRUE,
+    overwrite = overwrite
+  )
+
+  use_rextendr_template(
+    "configure.win",
+    save_as = file.path("configure.win"),
+    quiet = quiet,
+    as_executable = TRUE,
+    overwrite = overwrite
+  )
+
+  use_rextendr_template(
     "entrypoint.c",
     save_as = file.path("src", "entrypoint.c"),
     quiet = quiet,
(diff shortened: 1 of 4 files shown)


R/use_extendr.R
```diff
@@ -265,6 +283,10 @@ use_rextendr_template <- function(template,
     overwrite = overwrite
   )
 
+  if (isTRUE(as_executable)) {
+    Sys.chmod(save_as, mode = "755")
```
eitsupi, line 287: I do not know if this will work on Windows.

eitsupi: It seems working on CI

R/use_extendr.R
```diff
@@ -68,6 +68,22 @@ use_extendr <- function(path = ".",
...
+  )
+
+  use_rextendr_template(
+    "configure.win",
+    save_as = file.path("configure.win"),
+    quiet = quiet,
+    as_executable = TRUE,
+    overwrite = overwrite
+  )
+
```
JosiahParry, line 86: I wonder if it would make more sense to have the Sys.chmod() calls separated out for these or included in the `use_rextendr_template()`. I think this current change is actually quite nice because who knows what will need to be executable in the future.

R/use_extendr.R
JosiahParry: It would be good to have a windows user confirm this. Given that it is part of `{base}` I think it ought to work. How it works may be different, though?

> It may not be supported on a system (when a warning is issued).

So long as there is no warning emitted in CI, I think we can confidently say it works on windows.

JosiahParry:
This change would be another step towards making it easier to get packages on CRAN—stil harder than it ought to. 

This PR looks cut and dry to me and would be an overall win.

Ilia-Kosenkov (changes requested):
Also I think you need to pull in changes from `main` to resolve `lintr` issues.

PR #311, opened 2023-08-15
feat: An exprimental function for vendoring Rust crates

### LINKED ISSUE #300: Tools to build a package with all dependent crates bundled?
From https://github.com/pola-rs/r-polars/pull/308#issuecomment-1632405677
Ref: https://cran.r-project.org/web/packages/using_rust.html, https://github.com/yutannihilation/string2path, https://yutani.rbind.io/post/rust-and-cran-repository-policy/

CRAN announced the following as of a few days ago.

> So downloading of Rust ‘crates’ will in future require the explicit permission of the CRAN team.
>
> A better alternative is to bundle all the Rust software in the package. Thanks to Hiroaki Yutani for providing an example of doing that in his string2path package (version 0.1.5 at the time of writing). Even if the bundle is too large to include in the package sources, it could be downloaded from elsewhere (and check-summed). 

Currently the extendr template does not include dependent crates in the source package and assumes that the user downloads them from the Internet.
It would be great if an easy way could be provided to build a package with all the dependency crates bundled.

`string2path` commits all dependent crates to Git as a single compressed file, but I think it is better to have the bundled crates gitignored if possible, since this will cause constant diffs of huge files.

eitsupi:
Close #300

This makes no sense unless used in combination with the contents of `Makevars`, so there is probably little value in calling this function directly.

Would it be better to control the behavior with something like `Config/rextendr/vendoring: true`?

### DIFF
--- a/R/vendoring.R
+++ b/R/vendoring.R
@@ -1,0 +1,55 @@
+#' Vendor dependent Rust crates
+#'
+#' Vendoring all dependent Rust crates into a single tarball.
+#' @inheritParams use_extendr
+#' @return No return value, called for side effects.
+#' @noRd
+vendor_crates <- function(path = ".") {
+  withr::with_dir(path, usethis::use_build_ignore("src/rust/vendor"))
+
+  src_dir <- rprojroot::find_package_root_file("src", path = path)
+
+  out_file <- file.path(src_dir, "rust", "vendor.tar.xz")
+  config_toml_file <- file.path(src_dir, "rust", "vendor-config.toml")
+
+  vendor_rel_path <- file.path("rust", "vendor")
+
+  withr::local_dir(src_dir)
+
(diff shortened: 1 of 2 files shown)


eitsupi:
(There is no test on this yet, but I have changed the status for review.)

R/vendoring.R
```diff
@@ -0,0 +1,55 @@
+#' Vendor dependent Rust crates
+#'
+#' Vendoring all dependent Rust crates into a single tarball.
+#' @inheritParams use_extendr
+#' @return No return value, called for side effects.
+#' @noRd
```
CGMossa, line 6: Curious, why `@noRd`?

eitsupi: > Curious, why `@noRd`?

As I said in the comments, I don't think it's worth running this function by itself.

PR #312, opened 2023-08-15
fix(template): set `CARGO_BUILD_JOBS=2` on CRAN

eitsupi:
The CRAN policy limits the number of cores that can be used during installation to 2.

From <https://cran.r-project.org/web/packages/policies.html>

> Checking the package should take as little CPU time as possible, as the CRAN check farm is a very limited resource and there are thousands of packages. Long-running tests and vignette code can be made optional for checking, but do ensure that the checks that are left do exercise all the features of the package.
>
> If running a package uses multiple threads/cores it must never use more than two simultaneously: the check farm is a shared resource and will typically be running many checks simultaneously. 

From <https://cran.r-project.org/web/packages/using_rust.html>

> cargo build -j N defaults to the number of ‘logical CPUs’. This usually exceeds the maximum allowed in the CRAN policy, so needs to be set explicitly to N=1 or 2.

### DIFF
--- a/inst/templates/Makevars
+++ b/inst/templates/Makevars
@@ -15,6 +15,7 @@ $(STATLIB):
 	# therefore is only used if cargo is absent from the user's PATH.
 	if [ "$(NOT_CRAN)" != "true" ]; then \
 		export CARGO_HOME=$(CARGOTMP); \
+		export CARGO_BUILD_JOBS=2; \
 	fi && \
 		export PATH="$(PATH):$(HOME)/.cargo/bin" && \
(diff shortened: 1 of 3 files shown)


eitsupi:
Note that hellorust switched from configuration by `CARGO_BUILD_JOBS=2` to using the `-j 2` option between 1.1.0 and 1.1.1.
https://github.com/r-rust/hellorust/commit/a3ae6c24d062225e40607480386eaaaed6cddeb1

The reason is unknown.

I am not sure if using `CARGO_BUILD_JOBS=2` here is a bad idea, as either setting should have the same effect.

PR #313, opened 2023-08-30
Adds CRAN compliant vendoring helper functions

JosiahParry:
This PR adds functionality to vendor rust dependencies within an R package. 
It incorporates much of the draft PR #311. 

It adds 2 functions 

- `use_cargo_vendor()`: creates the package scaffolding for using `cargo vendor` in a CRAN compliant way. It does so by:
  - modifying `Makevars` and `Makevars.win` to compile with only 2 cores, always compile in offline mode, and print the versions of rustc and cargo
  - adds a template `vendor-config.toml` file
  - adds the appropriate files to both `.Rbuildignore` and `.gitignore` 
- `vendor_pkgs()` is derived from PR #311. It 
  - runs `cargo vendor` to create the `vendor` directory
  - checks if the suggested `config.toml` differs from `vendor-config.toml` and if so, it updates it accordingly
  - compresses the `vendor` directory into a `vendor.tar.xz`
  - invisibly returns a data frame with two columns containing the crate name and the version of the crate 

The end user experience is this:

- `rextendr::use_cargo_vendor()` to instantiate vendoring
- `rextendr::vendor_pkgs()` to build their compressed zip file or update it

### Follow up tasks: 

- submit a package to CRAN using this workflow. I plan on doing this for rsgeo given that the [r-universe build worked without any problems](https://github.com/r-universe/josiahparry/actions/runs/6027293671/job/16352313498) using these functions
- upon successful CRAN submission, document this workflow in an rextendr vignette

### Question for reviewers

There is one question, though, based on #312. Right now the template Makevars add `-J 2 --offline` to the cargo build statement. Should the templates use the `CARGO_BUILD_JOBS` environment variable instead? 

### Design notes

I opted to not modify the existing Makevars and Makevars.win file to use offline and -J 2 options in an if statement. This is because i think it would lead to an even more complex Makevar file which would build differently based on the presence of CRAN or not.

Additionally, I think the vendoring experience should be opt in and done knowingly.


DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: rextendr
 Title: Call Rust Code from R using the 'extendr' Crate
-Version: 0.3.1.9000
+Version: 0.4.0
```
JosiahParry, line 3: this was accidental! Will fix

JosiahParry: Aye yi yi got the version wrong again! NOW its fixed

Ilia-Kosenkov:
This is beyond awesome! To begin with, can you walk me through the process of creating a fresh dummy package and use this functionality? I want to test it locally. Is it enough to just run these two methods one by one?

Another important thing would be to test this workflow on CI. We have a separate CI that simulates creating new package using `{rextendr}`. We need to have the same but testing vendor capability.

Ilia-Kosenkov:
Also you can run `styler::style_pkg()` to please `{lintr}`

JosiahParry:
@Ilia-Kosenkov 

The process would go like this: 

```r
usethis::create_package()
rextendr::use_extendr()
rextendr::use_cargo_vendor()
```
do some stuff write some functions do some documenting etc.
Then when youre ready to compress the vendored packages you write: 

```r
rextendr::vendor_pkgs()
```

Ilia-Kosenkov:
Ah and I think for starters you can write a snapshot test to see what files are generated when you run this. Check out existing snapshot tests on `rextendr::use_extendr()`. You can track your test coverage here https://app.codecov.io/github/extendr/rextendr/pull/313

JosiahParry:
Awesome! I'll get to that in a few hours or first thing in the morning tomorrow.

Ilia-Kosenkov:
You have a typo somewhere
```
 Warning: Missing or unexported object: ‘cli::ci_abort’
 ```

**Comment by Ilia-Kosenkov** [comment]
Ok, vendoring works fine locally. What do I do next? Can I just install this package from my directory or what?

**Comment by JosiahParry** [comment]
@Ilia-Kosenkov correct! Once vendoring works you can `devtools::install()` successfully. Publishing to `r-universe` should also work without a hitch. 

I think this should be part of an rextendr specific "release" process

**Comment by JosiahParry** [comment]
> 



> Ok, vendoring works fine locally. What do I do next? Can I just install this package from my directory or what? I figured it out. Works nicely, however there is a UX issue -- if you use `rextendr::use_extendr()` and then `rextendr::use_cargo_vendor()`, `Makevars` are not updated since `overwrite = FALSE` by default. Perhaps we should do something like `{usethis}` does when it asks whether we want to do something with some weird answers to choose from, like `Yes`, `Absolutely not!`, `No way`?

Good call!

**Comment by Ilia-Kosenkov** [comment]
Also this is worth adding to `Readme.Rmd` I believe. Just a small section explaining what it is and why we need that (if you have links to CRAN mailing lists with bundling demands, insert them there)

**Comment by CGMossa** [comment]
Pinging @eitsupi because they had a draft PR doing this as well. 
I hope you can give some insight to this approach.

**Comment by eitsupi** [comment]
What is the reason for prioritizing this over my some PRs?

**Comment by JosiahParry** [comment]
@eitsupi I apologize if it feels like im usurping you! I didn't notice the draft PR. I tossed out what I had for vendoring in favor of the the proposal that was included in the draft #311.

**Comment by JosiahParry** [comment]
I think i've ticked all the boxes! @Ilia-Kosenkov

**Comment by CGMossa** [comment]
> What is the reason for prioritizing this over my some PRs?

I don't think there is any reason. We have massive guilt due to neglecting this project.. Then a new PR drop and we think.. Let's just get this thing out.

I have many, many neglected PRs over at extendr project... 🤷‍♂️

**Comment by Ilia-Kosenkov** [comment]
Hey @eitsupi , sorry for the PR neglect, it was an error on our side (and my oversight personally since I was looking only at GH notifications instead of browsing all PRs). Since your PR is in a draft phase, I'd suggest abandoning it in favor of this one, with you reviewing & QA this PR since you already know what should be working.

**Comment by JosiahParry** [comment]
> Hm, can we also remove vendored crates when calling `rextendr::clean()`? I did not check it myself, but I believe it should remove this as it is an artifact .

I do not believe `cargo vendor` is capable of removing the entire `vendor` directory. You can use `cargo clean -p` to remove a package as you would without vendoring just fine. `rextendr::clean()` does not remove the vendored libraries. 

Please let me know what additional changes are requested.

**Comment by eitsupi** [comment]
I think #311 and this PR cover very different areas.
#311 only adds a function and does not mention any changes to the package development workflow.
The reason is as described in #311.

> I tossed out what I had for vendoring in favor of the the proposal that was included in the draft #311.

I have looked at this a bit and don't think it follows the tar command options.

> Since your PR is in a draft phase, I'd suggest abandoning it in favor of this one, with you reviewing & QA this PR since you already know what should be working.

My PR is a draft because I was looking for opinions. In real life I probably failed because no one saw my PR and I was not able to gather opinions.

**Comment by eitsupi** [comment]
@Ilia-Kosenkov Could you add @JosiahParry and I as external collaborators for this repository?
That way we can edit the PR directly.

**Comment by JosiahParry** [comment]
Rather than edit the PR directly, can we please make inline comments using the review feature to discuss it? All tar-ing is taken directly from both PR #311 and hellorust https://github.com/r-rust/hellorust/blob/8e5e619dff723708b8e7e99f5f679a6c71a85c43/src/myrustlib/vendor-update.sh#L4.

**Comment by eitsupi** [comment]
> Rather than edit the PR directly, can we please make inline comments using the review feature to discuss it? All tar-ing is taken directly from both PR #311 and hellorust [r-rust/hellorust@`8e5e619`/src/myrustlib/vendor-update.sh#L4](https://github.com/r-rust/hellorust/blob/8e5e619dff723708b8e7e99f5f679a6c71a85c43/src/myrustlib/vendor-update.sh#L4).

As far as I have tried, the command in hellorust changes the hash of the tarball each time it is run.
#311's one is based on the command used by string2path and produces the same tarball each time (See https://github.com/eitsupi/prqlr/pull/152#issuecomment-1652759510), with the highest compression ratio.

File: R/vendor.R [hunk 2]
```diff
@@ -0,0 +1,184 @@
...
+  if (!configs_match) {
+    brio::write_lines(config_toml, file.path(src_dir, "vendor-config.toml"))
+    cli::cli_alert_info("Updating {.file src/rust/vendor-config.toml}")
+  }
+
+  # compress the vendor.tar.xz
+  compress_res <- withr::with_dir(src_dir, {
+    processx::run(
+      "tar", c(
+        "-cJ", "--no-xattrs", "-f", "vendor.tar.xz", "vendor"
```
JosiahParry, line 161: See: https://github.com/extendr/rextendr/pull/313#issuecomment-1700908011
Current behavior changes the hash (which the test is based on).
Use args from #311 to ensure higher compression and consistent hashing.

README.Rmd
```diff
@@ -80,8 +80,8 @@ rust_function(
           Either::Right(x) => Either::Right(x.iter().sum()),
       }
   }",
-  use_dev_extendr = TRUE,                        # Use development version of extendr from GitHub
-  features = "either",                           # Enable support for Either crate
+  use_dev_extendr = TRUE, # Use development version of extendr from GitHub
+  features = "either", # Enable support for Either crate
```
JosiahParry, line 84: these are from styler

R/vendor.R
JosiahParry: I have confirmed hashing behavior.

JosiahParry: I realized that the R-core team has already made a utility to make this easier 🤦🏼 
`tar("../vendor.tar.xz", compression = "xz", compression_level = 9)`

eitsupi: I think I even tried the `tar` function at the time, but for some reason gave up using it.

Perhaps rextendr should test that the tarball hash is immutable.

JosiahParry: I agree. Its in this test
https://github.com/extendr/rextendr/blob/8ac7cf00ccb5e8ffea1e653ea8162170c6b18d10/tests/testthat/test-use_cargo_vendor.R#L30

inst/templates/vendor-config.toml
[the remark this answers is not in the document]
JosiahParry: As part of the build process in the Makevars it is copied to `.cargo/config.toml`. It is however updated when using `vendor_pkgs()` and might be redundant as such.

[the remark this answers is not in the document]
eitsupi: Ah, perhaps an empty file would suffice?

JosiahParry:
My goal is to make R + Rust be a viable future. As such I made this PR to be a thorough attempt at addressing the most recent CRAN guidance. This PR is tailored _strictly_ towards appeasing CRAN no more no less. 

I think a more appropriate place to discuss PRs #311 and #312 would be in those threads repsectively.

yutannihilation:
I think it's a reasonable choice. I personally don't use rextendr's templates because I need more opinionated ones for various reasons. Honestly, I don't think it's a good time to define "the standard" templates because the requirements change constantly and rapidly.

JosiahParry:
I agree! My goal here is to have a minimal template that can get folks less experienced than you both (such as myself) up and running without much thought. It is based on the prior art developed by @eitsupi, @yutannihilation, and Joeren. Your feedback and comments are greatly appreciated!

yutannihilation:
I appreciate your effort, but, in my opinion, "a minimal template" should not target CRAN.

JosiahParry:
> I appreciate your effort, but, in my opinion, "a minimal template" should not target CRAN.

To clarify, these functions are opt-in and not a replacement of the default templates.

R/vendor.R
JosiahParry: The `tar()` function is very weird....it expands in a very weird way. I've reverted so it will compile—i know it will work on ubuntu, windows, and mac. This can be improved in a follow up PR.

PR #319, opened 2023-09-17
[CI] Enable code cov status checks and aim at 70% of coverage

Ilia-Kosenkov:
I want to try out getting more info on code coverage in PRs.


Ilia-Kosenkov:
Seems current codecov status checks are enough, closing this for now.

JosiahParry (approved):
In general, I'm happy to have codecov be useful so that we don't merge untested R code. This will prevent any future accidental breaks. I don't hope that this doesn't result in more snapshot tests though! 

As I understand it, this PR requires that any PR have have least 70% code coverage. I'm quite okay with this. And if it will help us get to releasable states, I think it is good!

PR #321, opened 2023-11-17
extendr project template

kbvernon:
Trying to add required content to support RStudio project templates. Currently getting the following error: 
```r
Error: No root directory found in G:/My Drive or its parent directories. Root criterion: contains a file "DESCRIPTION" with contents matching "^Package: "
```


Ilia-Kosenkov:
Found an issue, and I will fix it in a separate PR.

kbvernon:
I believe I have this working, but I'm not sure what the M.O. is for doing tests.

kbvernon:
@Ilia-Kosenkov, do you have a recommendation for how to handle missing `usethis` before `use_extendr()` gets called? should it just error?

kbvernon:
`use_extendr()` uses `rlang::check_installed("usethis")`. However, that opens an interactive context I don't think is available for project directory creation through the RStudio GUI. Maybe I should use a combination of `rlang::is_installed("usethis")` and `cli::abort()` instead?

Ilia-Kosenkov:
You can also use this in your function, call it before you access anything `{usethis}` - related.
https://github.com/extendr/rextendr/blob/09bfebae90e679cbb3a21e9ce983c032c82e1828/R/use_extendr.R#L38

R/create_extendr_package.R
```diff
@@ -0,0 +1,76 @@
...
+
+  # error if usethis is not installed
+  rlang::check_installed("usethis")
+
+  args <- rlang::list2(...)
+
+  # hunch is that rstudio project text input widgets return empty strings
+  # when no value is given, want to make sure it is NULL so `use_extendr()`
+  # handles it correctly
+  args <- lapply(args, function(x) if (x == "") return(NULL) else return(x))
```
Ilia-Kosenkov, line 26: We have `{purrr}` as dependency, so you can use that. Also, `==` will backfire horribly if `x` is `NA` or `NULL`

kbvernon: I believe these widgets are based on shiny, so the text input widget should return `""` when empty. https://shiny.posit.co/r/reference/shiny/1.7.3/textinput. 

but even so, maybe we shouldn't rely on that assumption? should i do something like `if (is.na(x) | is.null(x) | nchar(x) == 0)`?

kbvernon: `rlang` to the rescue!

R/create_extendr_package.R
```diff
@@ -0,0 +1,80 @@
...
+#'
+#' @param path a path to new directory
+#' @param ... arguments passed on to `usethis::create_package()` and
+#' `rextendr::use_extendr()`
+#'
+#' @return Path to the newly created project or package, invisibly.
+#' @keywords internal
+#'
+#' @examples
+create_extendr_package <- function(path, ...) {
```
Ilia-Kosenkov, line 16: Should we mark it as `@noRd`? @CGMossa what's your take?

R/create_extendr_package.R
```diff
@@ -0,0 +1,80 @@
...
+  # hunch is that rstudio project text input widgets return empty strings
+  # when no value is given, want to make sure it is NULL so `use_extendr()`
+  # handles it correctly
+  nullify_empty_string <- function(x) {
+    if (rlang::is_string(x) && nzchar(x)) x else NULL
+  }
+
+  args <- purrr::map(args, nullify_empty_string)
+
+  # build package directory, but don't open yet!
```
Ilia-Kosenkov, line 32: What does 'open' mean here?

R/create_extendr_package.R
```diff
@@ -0,0 +1,80 @@
...
+    ""
+  )
+
+  text <- c(
+    "NOTE:",
+    "To use {rextendr} in any meaningful way, the user must have",
+    "Rust and Cargo available on their local machine. To check that you do,",
+    "please run `rextendr::rust_sitrep()`. This will provide a",
+    "detailed report of the current state of your Rust infrastructure, along",
+    "with some helpful advice about how to address any issues that may arise."
```
Ilia-Kosenkov, line 67: Should we keep using the same personal pronouns here? If you start with 'the user', let's keep it like this, without switching to 'you'. How does this sound?

R/create_extendr_package.R
```diff
@@ -0,0 +1,80 @@
...
+  )
+
+  content <- paste(
+    paste(header, collapse = "\n"),
+    paste(text, collapse = "\n"),
+    sep = "\n"
+  )
+
+  writeLines(content, con = file.path(path, "INDEX"))
```
Ilia-Kosenkov, line 76: An oversight on my side, but could you please use `brio::write_lines()` here? We use it throughout the package (and other `{brio}` functions when necessary).

R/create_extendr_package.R
kbvernon: If it's internal, will roxygen still generate a documents page unless we explicitly use `@noRd`?

R/create_extendr_package.R
kbvernon: That's a parameter in `usethis::create_package()`. Evidently, it can mean several different things depending on the context, but it basically means "open the new project in a new R session or change the active project in RStudio to the new one." I believe we want to set it to FALSE and let the RStudio project GUI handle that instead. But, do you want me to add the text from the [{usethis} docs](https://usethis.r-lib.org/reference/create_package.html) to that comment for clarification?

R/create_extendr_package.R
kbvernon: I agree! And appreciate the level of consistency. Will change this today!

R/create_extendr_package.R
Ilia-Kosenkov: Just check it, did you run `devtools::document()`?

R/create_extendr_package.R
Ilia-Kosenkov: Can you just rephrase it so it is better understood by people not involved with RStudio?

kbvernon: Yes! Will do!

R/create_extendr_package.R
kbvernon: I did not! Is that something you want all contributors to do for each pull request? Should I also do `devtools::check()`?

Ilia-Kosenkov: Document - yes, since we do not commit changes on CI. Check - up to you, but this is done on CI and it is ✅

kbvernon:
Per our Discord discussion, I added commits so that `create_extendr_package()` no longer adds a temporary INDEX file to the project folder. This made the single `{testthat}` test irrelevant, so I also reverted that commit.

R/create_extendr_package.R
kbvernon: this should be done with recent commit, let me know if that is sufficient

R/create_extendr_package.R
kbvernon: no longer relevant with recent commit

R/create_extendr_package.R
kbvernon: no longer relevant with recent commit

R/create_extendr_package.R
```diff
@@ -0,0 +1,55 @@
+
+#' Create a project for R package development with Rust
```
JosiahParry, line 2: Consider rewriting `Create a package using extendr` or `Create package that uses Rust` or something. 

Generally, I dislike "for R package development." We're writing R so thats a given.

R/create_extendr_package.R
```diff
@@ -0,0 +1,55 @@
...
+#'
+#' @description
+#' This function creates an R project directory for package development
+#' with Rust extensions.
+#'
+#' @param path a path to new directory
```
JosiahParry, line 8:
```suggestion
@inheritParams usethis::create_package
```

R/create_extendr_package.R
```diff
@@ -0,0 +1,55 @@
...
+#' @param ... arguments passed on to `usethis::create_package()` and
+#' `rextendr::use_extendr()`
+#'
+#' @return Path to the newly created project or package, invisibly.
+#' @keywords internal
+#'
+#' @noRd
```
JosiahParry, line 15: Is there any reason why we wouldn't export this function? Seems like a handy function to be exported.

R/create_extendr_package.R
```diff
@@ -0,0 +1,55 @@
...
+  use_extendr(
+    path,
+    crate_name = args[["crate_name"]],
+    lib_name = args[["lib_name"]],
+    quiet = TRUE,
+    overwrite = TRUE,
+    edition = args[["edition"]] %||% TRUE
+  )
+
+  return(invisible(path))
```
JosiahParry, line 53: you do not need to explicitly call return here since `invisible(path)` is the last thing included. Reserve `return()` for early function returns.

JosiahParry (changes requested):

R/create_extendr_package.R
kbvernon: I believe that's a question for you and @Ilia-Kosenkov to decide on.

JosiahParry: I'd vote for @export here to include it in the namespace. But that's not necessary for this PR.

JosiahParry:
Package is created and opened interactively with RStudio. The infrastructure works as well. This should not be exported until the arguments are changed and handled differently. 

This is because you're capturing all arguments via `...` and _expecting_ there to be a valid value passed in. This is true when using the wizard. But it is _not_ true when doing interactively. The highlighted lines are where the problem comes from. 

https://github.com/extendr/rextendr/blob/cbdb5b6f70b0db27f3ec97123883231f8af308f5/R/create_extendr_package.R#L34-L52
image

JosiahParry:
Since this is not an exported function I am comfortable merging.

JosiahParry (approved):

PR #323, opened 2023-11-19
[Bugfix] Correctly handle `use_extendr(custom_path)`

Ilia-Kosenkov:
Correctly handle situations when `use_extendr()` is called with a non-default path. Now it temporarily sets the working directory and `{usthis}` path to the package root.

### DIFF
--- a/R/use_extendr.R
+++ b/R/use_extendr.R
@@ -37,14 +37,30 @@ use_extendr <- function(path = ".",
 
   rlang::check_installed("usethis")
 
-  rextendr_setup(path = path)
-
-  root_path <- normalizePath(rprojroot::find_package_root_file(path = path), mustWork = FALSE, winslash = "/")
+  # Root path computed from user input
+  root_path <- try_get_root_path(path)
+  # Root path computed from `{usethis}`
   usethis_proj_path <- try_get_proj_path()
 
+  # If they do not match, something is off, try to set up temporary project
   if (!isTRUE(root_path == usethis_proj_path)) {
     usethis::local_project(path, quiet = quiet)
   }
+
+  # Check project path once again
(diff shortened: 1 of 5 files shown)


R/use_extendr.R
```diff
@@ -37,9 +37,32 @@ use_extendr <- function(path = ".",
 
   rlang::check_installed("usethis")
 
-  rextendr_setup(path = path)
+  # Root path computed from user input
+  root_path <- try_get_root_path(path)
+  # Root path computed from `{usethis}`
+  usethis_proj_path <- try_get_proj_path()
+
+  # If they do not match, something is off, try to set up temporary project
```
CGMossa, line 45: what? what's going on here?

CGMossa (approved):
LGTM, but I don't get this setting up of a pseudo project..

R/use_extendr.R
Ilia-Kosenkov: So the code that we run depends on two things - path that we get from the user (impacts direct writes or rprojroot) and active usethis project (impacts all calls to usethis). If one of these two is out of whack, something will not work. Now in the easiest case of creating new package and immediately running use_extendr, it always works. But if your run use_extendr later or from a different folder, we need to adjust for that. We do this with temporary path and project, pretending we are in a correct folder to execute everything, and then we restore the original project/path configuration.

CGMossa: seems odd, but to be fair, I have alerted to similar issues elsewhere in the code. As always, my policy is to merge what works reliable, and improve later on. godspeed @Ilia-Kosenkov !

Ilia-Kosenkov:
Checked against #321, it seems to pass `use_extendr()` step. Merging now 
image

PR #337, opened 2024-01-27
add roxygen export tags to methods

kbvernon:
Recent updates to Roxygen suggest adding export tags to all methods even if you're not exporting the generic. From their NEWS.md: 

> The NAMESPACE roclet now reports if you have S3 methods that are missing an \@ export tag. All S3 methods need to be \@ exported (which confusingly really registers the method) even if the generic is not. This avoids rare, but hard to debug, problems (#1175). You can suppress the warning with @exportS3Method NULL (#1550).

So, I did this for all the methods in _toml_serialization.R_.


Ilia-Kosenkov (approved):
LGTM, @kbvernon  can I merge this?

kbvernon:
Thanks! Yes, if you're ready for it to go, then I am!

PR #343, opened 2024-03-10
[#342] Display warning when `NAMESPACE` does not contain `useDynLib()`

### LINKED ISSUE #342: Issue when creating packages through the RStudio GUI
Hello!

I came accross an issue when creating packages through the RStudio GUI. When we create a "clean package" like this: 
image

The NAMESPACE file created will not be written by roxygen2 and comes out like this:

image

Causing an issue when invoking `rextendr::document()` which will make it skip writing the NAMESPACE file:

```
─  DONE (extendr)
✔ Writing 'R/extendr-wrappers.R'
ℹ Updating extendr documentation
ℹ Loading extendr
✖ extendr-wrappers.R:12: `@docType "package"` is deprecated.
ℹ Please document "_PACKAGE" instead.
✖ Skipping NAMESPACE
ℹ It already exists and was not generated by roxygen2.

```

This in turn will make it so the wrap objects are not exported when installing the package which will cause the error when trying to call the function:

`Error in odsrust::hello_world() : object 'wrap___hello_world' not found`

When creating the package using `usethis::create_package(".")` or the new rextendr project template this problem doesn't happen, due to the NAMESPACE file being written corretly by roxygen2 such as in the example below:

image

Hopefully this can help someone with the same issue as I had.

Thanks and congratulations on the project!

Ilia-Kosenkov:
Fixed #342

Add explicit warning about missing `useDynLib()` from `NAMESPACE`.
image


CGMossa (approved):
LGTM

Issue #360, opened 2024-06-18
FR: rextendr::use_crate()?


kbvernon:
Similar to `usethis::use_package()` but for adding dependencies to Cargo.toml.

Happy to contribute a PR for this if you all think it is worthwhile.

JosiahParry:
Yes definitely! You should definitely do this. We do not need to have all the arguments to the CLI either. Just the most common ones. If you need things like optional or nightly etc then you probably wont need an R package to help! 

Here's some start code we can use: 

```rs
compact <- function(.x) Filter(length, .x)

use_crate <- function(crate, features = NULL, git = NULL, path = ".") {
  # TODO check arguments and that they are the right type and length
  # consider adding standalone type checks 

  # get source directory
  src_dir <- rprojroot::find_package_root_file(path, "src/rust")

  # craft additional args
  cargo_add_opts <- compact(list(
    "--features" = features,
    "--git" = git
  ))

  # craft the additional arguments
  adtl_args <- unname(unlist(Map(
    function(.x, .nm) {
      paste(.nm, paste0(.x, collapse = " "))
    },
    cargo_add_opts,
    names(cargo_add_opts)
  )))
  
  # run the commmand
  processx::run(
    "cargo",
    c("add", crate, adtl_args),
    echo_cmd = TRUE,
    wd = src_dir
  )

}
```

kbvernon:
Thanks, @JosiahParry! I'll try to get a PR together this evening. 

But also, what about adding an `options` parameter that would take a named list to catch all the other flags?

JosiahParry:
my preference would be to use `...` and use `rlang::dots_list()` to capture the others!

Ilia-Kosenkov:
@kbvernon , did your PR resolve this issue or are you planning on adding more features?

kbvernon:
Ah, yeah, this issue is resolved with https://github.com/extendr/rextendr/pull/361. Sorry, I forgot to close it!

PR #361, opened 2024-06-19
draft of use_crate()

kbvernon:
For https://github.com/extendr/rextendr/issues/360

### DIFF
--- a/R/use_crate.R
+++ b/R/use_crate.R
@@ -46,90 +46,15 @@ use_crate <- function(
     path = "."
 ){
 
-  # check crate
-  if (!rlang::is_character(crate) && length(crate) != 1){
+  # check args
+  check_string(crate)
+  check_character(features, allow_null = TRUE)
+  check_string(git, allow_null = TRUE)
+  check_string(version, allow_null = TRUE)
+  check_bool(optional)
+  check_string(path)
 
-    info <- paste(
-      "You supplied an object of class {class(crate)[1]}",
-      "with length {length(crate)}."
-    )
-
-    cli::cli_abort(
-      "{.var crate} should be a length one character vector.",
-      "i" =  info,
-      class = "rextendr_error"
-    )
-
-  }
-
-  # check features
-  if (!is.null(features) && !rlang::is_character(features)){
-
-    cli::cli_abort(
-      "{.var features} should be a character vector.",
-      "i" = "You supplied an object of class {class(features)[1]}.",
-      class = "rextendr_error"
-    )
-
-  }
-
-  # check git
-  if (!is.null(git)){
-
-    if (!rlang::is_character(git) && length(git) != 1){
-
-      info <- paste(
-        "You supplied an object of class {class(git)[1]}",
-        "with length {length(git)}."  
-      )
-
-      cli::cli_abort(
-        "{.var git} should be a length one character vector.",
-        "i" = info,
-        class = "rextendr_error"
-      )
-
-    }
-
-  }
-
-  # check version
-  if (!is.null(version)){
-
-    if (!rlang::is_character(version) && length(version) != 1){
-
-      info <- paste(
-        "You supplied an object of class {class(version)}",
-        "with length {length(version)}."
-      )
-
-      cli::cli_abort(
-        "{.var version} should be a length one character vector.",
-        "i" = info,
-        class = "rextendr_error"
-      )
-
-    }
-
-    crate <- paste0(crate, "@", version)
-
-  }
-
-  #check optional
-  if (!rlang::is_bool(optional) && length(optional) != 1){
-
-    info <- paste(
-      "You supplied an object of class {class(optional)[1]}",
-      "with length {length(optional)}."
-    )
-
-    cli::cli_abort(
-      "{.var optional} should be a length one boolean vector.",
-      "i" = info,
-      class = "rextendr_error"
-    )
-
-  }
+  if (!is.null(version)){ crate <- paste0(crate, "@", version) }
 
   # combine main options
   cargo_add_opts <- list(
@@ -139,7 +64,7 @@ use_crate <- function(
   )
 
   # clear empty options
-  cargo_add_opts <- Filter(length, cargo_add_opts)
+  cargo_add_opts <- purrr::discard(cargo_add_opts, rlang::is_empty)
 
   # combine option names and values into single strings
   adtl_args <- unname(purrr::imap_chr(
--- a/R/use_crate.R
+++ b/R/use_crate.R
@@ -5,7 +5,7 @@
 #' @param crate character scalar, the name of the crate to add
 #' @param features character vector, a list of features to include from the
 #' crate
-#' @param git character scalar, the URL of the Github repository
+#' @param git character scalar, the full URL of the remote Git repository
 #' @param version character scalar, the version of the crate to add
 #' @param optional boolean scalar, whether to mark the dependency as optional 
 #' (FALSE by default)
@@ -29,7 +29,7 @@
 #' use_crate("serde", features = "derive")
 #'
 #' # add to [dependencies] using github repository as source
-#' use_crate("serde", git = "serde-rs/serde")
+#' use_crate("serde", git = "https://github.com/serde-rs/serde")
 #'
 #' # add to [dependencies] with specific version
 #' use_crate("serde", version = "1.0.1")
@@ -43,8 +43,7 @@ use_crate <- function(
     git = NULL,
     version = NULL,
     optional = FALSE,
-    path = ".",
-    ...
+    path = "."
 ){
 
   # check crate
@@ -91,8 +90,6 @@ use_crate <- function(
       )
 
     }
-
-    git <- paste0("https://github.com/", git)
 
   }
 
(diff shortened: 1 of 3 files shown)


R/use_crate.R
```diff
@@ -0,0 +1,141 @@
...
+    )
+
+  }
+
+  # check features
+  if (!is.null(features) && !rlang::is_character(features)){
+
+    cli::cli_abort(
+      "{.var features} should be a character vector.",
+      "i" = "You supplied an object of class {class(features)}.",
```
JosiahParry, line 60: this `class()` call should grab the first element of the output. There may be 1 or more classes

R/use_crate.R
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  # check git
+  if (!is.null(git)){
+
+    if (!rlang::is_character(git) && length(git) != 1){
+
+      cli::cli_abort(
+        "{.var git} should be a length one character vector.",
+        "i" = "You supplied an object of class {class(git)} with length {length(git)}.",
```
JosiahParry, line 73: same thing here

R/use_crate.R
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  # check version
+  if (!is.null(version)){
+
+    if (!rlang::is_character(version) && length(version) != 1){
+
+      cli::cli_abort(
+        "{.var version} should be a length one character vector.",
+        "i" = "You supplied an object of class {class(version)} with length {length(version)}.",
```
JosiahParry, line 90: same thing

R/use_crate.R
```diff
@@ -0,0 +1,141 @@
...
+  # combine main options
+  cargo_add_opts <- list(
+    "--features" = features,
+    "--git" = git
+  )
+
+  # add additional options from ...
+  lst <- rlang::dots_list(...)
+
+  if (length(lst) > 0){ names(lst) <- paste0("--", names(lst)) }
```
JosiahParry, line 109: this part might be odd. this would probably need to be better described in the params and with an example. like `optional = TRUE` becomes `--optional true`

R/use_crate.R
kbvernon: good point. are you thinking `class(x)[1]` or is there a better way?

R/use_crate.R
kbvernon: there are also flags that don't take a value, like `--dev`. i wasn't sure how to handle those...

R/use_crate.R
JosiahParry: that works for me! We could also utilize the `obj_type_friendly()` standalone function.
https://github.com/r-lib/rlang/blob/main/R/standalone-obj-type.R

R/use_crate.R
JosiahParry: I'm not sure we need to. We can just be very specific and provide two options: `features` and `optional` anything else they can just go into the terminal and handle it themselves.

kbvernon: makes sense to me! give me one second.

R/use_crate.R
```diff
@@ -0,0 +1,172 @@
...
+      cli::cli_abort(
+        "{.var git} should be a length one character vector.",
+        "i" = info,
+        class = "rextendr_error"
+      )
+
+    }
+
+    git <- paste0("https://github.com/", git)
+
```
JosiahParry, line 96: i think we should expect everyone to provide the git path in its entirety instead of just `user/repo`. There is a chance they use bitbucket or gitlab etc.

R/use_crate.R
```diff
@@ -0,0 +1,172 @@
...
+#' use_crate("serde", optional = TRUE)
+#' }
+use_crate <- function(
+    crate,
+    features = NULL,
+    git = NULL,
+    version = NULL,
+    optional = FALSE,
+    path = ".",
+    ...
```
JosiahParry, line 47: i think we could probably delete this for now

R/use_crate.R
kbvernon: yeah, that's a solid point

R/use_crate.R
```diff
@@ -0,0 +1,172 @@
...
+      "{.var optional} should be a length one boolean vector.",
+      "i" = info,
+      class = "rextendr_error"
+    )
+
+  }
+
+  # combine main options
+  cargo_add_opts <- list(
+    "--features" = features,
```
JosiahParry, line 139: features will need to be specified as a space separated value 
```suggestion
    "--features" = paste0(features, collapse = " "),
```

JosiahParry: Erm i guess this should only happen if the length of features > 1

R/use_crate.R
```diff
@@ -0,0 +1,172 @@
...
+    "--optional" = tolower(as.character(optional))
+  )
+
+  # clear empty options
+  cargo_add_opts <- Filter(length, cargo_add_opts)
+
+  # combine option names and values into single strings
+  adtl_args <- unname(purrr::imap_chr(
+    cargo_add_opts,
+    \(x, i){ paste(i, paste0(x, collapse = " ")) }
```
JosiahParry, line 150: you can't use this shorthand since it requires we set a minimum version of R to 4.1. 
```suggestion
    function(x, i){ paste(i, paste0(x, collapse = " ")) }
```

R/use_crate.R
```diff
@@ -0,0 +1,169 @@
...
+        "i" = info,
+        class = "rextendr_error"
+      )
+
+    }
+
+  }
+
+  # check version
+  if (!is.null(version)){
```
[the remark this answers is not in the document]
kbvernon: Yeah, good point! But it raises the question: Is this validation logic something you'd also like to see for the whole package?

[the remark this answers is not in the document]
JosiahParry: I recommend using the standalone rlang checks. I am using these in all of my packages now. They only rely on R lang and make everything much much better to work with.

In the below I modified the existing standalones to use `rlang::` instead of the default which requires `#' @import rlang` 
https://github.com/R-ArcGIS/arcgisgeocode/blob/main/R/import-standalone-types-check.R
https://github.com/R-ArcGIS/arcgisgeocode/blob/main/R/import-standalone-obj-type.R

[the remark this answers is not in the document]
kbvernon: I'm happy to implement Josiah's suggestion, but I would need the go-ahead from one of you folks in charge. And there's also the question if we would want to implement these validation tools everywhere in the package.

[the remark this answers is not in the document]
Ilia-Kosenkov: You can put them in a separate file and start using here. We can slowly adopt em where applicable (out of scope of this PR).

Ilia-Kosenkov:
Do we need all two files, 400 lines each? Can we trim it only to used functions?

kbvernon:
Possibly, but if we intend to incrementally update other parts of rextendr to use these standalone checks, maybe we should think about which ones we are likely to use elsewhere (not just for this function) and then trim everything else. What do you think, @JosiahParry?

JosiahParry:
I think including the files is helpful for the future development of the package. I would suggest to keep them as is. 

This is because they add no other dependencies hence "standalone." Together they are 20kb which is very small. If size is a concern we can remove comments and whitespace. But again, this is a utility package, not one that is included in the final binary of another.

JosiahParry (approved):
🚀 Thank you for making this change. This alone will simplify the process of adding crates to one's `Cargo.toml` lowering the bar to entry for writing Rust powered R packages. 

If the standalone checks become an issue later while running checks prior to new version release we can revisit.

However, I think that they provide a significant QOL improvement that should not be overlooked.

CGMossa (approved):
I think it is fine as is right now.

I don't see a need to hand modify these functions and helpers to fit neatly in with this PR solely. It is a tool discovered by this PR, and should just be added if we like it.

kbvernon:
Awesome! Thanks! If you all want someone to work on integrating/modifying the standalone checks more, I can probably devote some time to it next month.

Ilia-Kosenkov:
@kbvernon , could you please also update `NEWS.md` since this is function is a part of public API?

kbvernon:
can do! thanks, @Ilia-Kosenkov!

R/use_crate.R
```diff
@@ -0,0 +1,94 @@
...
+
+  # combine main options
+  cargo_add_opts <- list(
+    "--features" = paste0(features, collapse = " "),
+    "--git" = git,
+    "--optional" = tolower(as.character(optional))
+  )
+
+  # clear empty options
+  cargo_add_opts <- purrr::discard(cargo_add_opts, rlang::is_empty)
```
JosiahParry, line 67: This change broke the function. Looks like we might be missing a test for this. I installed the new version 

```
rextendr::use_crate("lazy-static")
Running cargo add lazy-static '--features ' '--optional false'
Error:
! ! System command 'cargo' failed
```

This is because `"--features "` is included in the vector that is passed to processx. `--features`

kbvernon: yeah, i see this:
```r
purrr::discard(list(a = "", b = 1, c = NULL), rlang::is_empty)
# $a
# [1] ""
# 
# $b
# [1] 1
```

kbvernon: and 
```r
paste(NULL)
# character(0)
paste(NULL, collapse = " ")
# ""
```

PR #363, opened 2024-07-18
Fix lint

etiennebacher:
I was just testing `flint` on real packages and found this one, might as well fix it ;)


CGMossa:
Flint is already providing results!

CGMossa:
Rebased. Should work now as the error in CI was fixed in #365. Approving when test is passed, because we have auto-merge on.

Ilia-Kosenkov (approved):

PR #374, opened 2024-08-04
Support `r#` escape syntax in Rust function names in `rust_source()` family of functions

### LINKED ISSUE #356: some function with special name would not compile
for example, such rust code works fine:
```rust
fn r#try(a:i32, b:i32)->i32{a+b}
```
but rextendr has issue exporting it.

It is nothing more than an edge case. Things might be done with a hand-writting macro.

---

I'm trying writting a macro about auto exporting function without  explictly use the annoying macro:
```rust
extendr_module! {
    mod rext;
    fn hello_world;
}
```
Instead, if I success, just writting :
```rust
done!{Yes, it is done!};
```
is enough.

the trick should be re-define the function with the same name in a new module, for example, `_please_do_not_use_rmin_export_interface_as_your_mod_name_`, and use the following trick to merge all of the function in a single `mod`
```rust
mod private_foo {
    #[no_mangle]
    extern "C" fn foo(){} // not-so-private function
}
pub mod _please_do_not_use_rmin_export_interface_as_your_mod_name_ {
    pub extern "C" {
        fn foo();
    }
    const FOO_PTR:*const()=foo as *const_; // use pub extern fn foo::foo, which equals to the private_foo::foo thanks to the `#[no_mangle]` attribute.
}
```

---

That's all what I have found in the last week. Hope that may help develop a better rextendr:)

I might have several problem for packing R packages in the future (it cost me a day to switch the compile script to a ugly script compiles my crate), wondering is it possible to get some help here.

Ilia-Kosenkov:
Fixed #356


JosiahParry (approved):
👏🏼 Thank you! This is awesome. 🚀

PR #380, opened 2024-09-05
Address extendr-api CI failures

JosiahParry:
This PR addresses the failing CI tests in https://github.com/extendr/extendr/pull/842.

It does so by using the current head of the git repository for CI instead of the currently released version on crates.io. 

This also updates snapshots to pass the tests.

This should let us pass CI on extendr-api and prove that we can also patch extendr-api to keep up with the changes in libR-sys due to non-API. 

The question is: do we want to keep rextendr testing main? Or revert the change after being merged?

### DIFF
--- a/R/use_extendr.R
+++ b/R/use_extendr.R
@@ -137,7 +137,8 @@ use_extendr <- function(path = ".",
   cargo_toml_content <- to_toml(
     package = list(name = crate_name, publish = FALSE, version = "0.1.0", edition = edition),
     lib = list(`crate-type` = array("staticlib", 1), name = lib_name),
-    dependencies = list(`extendr-api` = "*")
+    dependencies = list(`extendr-api` = list(git = "https://github.com/extendr/extendr"))
+    # dependencies = list(`extendr-api` = "*")
   )
 
   use_rextendr_template(
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -25,7 +25,8 @@
 
     # Version of 'extendr_api' to be used
     rextendr.extendr_deps = list(
-      `extendr-api` = "*"
+      # `extendr-api` = "*"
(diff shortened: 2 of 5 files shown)


CGMossa:
Okay. I guess we should make an extendr patch release as soon as possible.

And rextendr clearly has some snapshot tests here that needs a cleanup.

Ilia-Kosenkov:
Why do you change defaults in {rextendr} though? If you need {rextendr} to test against main head, just configure tests so.

CGMossa:
> Why do you change defaults in {rextendr} though? If you need {rextendr} to test against main head, just configure tests so.

I'm not going to accept these changes and merge them, as I agree with you. But we are in a pinch, and integration test _with rextendr_ is just causing the extendr maintenance to be a nightmare. Atleast this PR shows that the integration test should pass..

I am also noticing @Ilia-Kosenkov that there are some snapshot tests in `rextendr` that needs to be updated here. 

I suggest that this PR is closed, but then it would be nice if @Ilia-Kosenkov can make a way to integrate test rextendr on extendr in a reasonable way, instead of this interlocking.

I also would recommend that we remove rextendr integration testing on extendr for now. It seems to me that the burden of maintaining the entire extendr ecosystem falls on people making PRs to extendr-api crate. Which is unreasonable. 

These are my thoughts atleast. Good job everyone.

Ilia-Kosenkov:
Ok, sounds fair. You can start by disabling rextendr tests in extendr for the time being. Feel free to mention me in any issues related to testing there, I might be able to help.

CGMossa:
I've takena  look again. I think there are some tests, that test extendr _on crates.io_ in rextendr. These should all be made, so they are disabled, or atleast follow the "default" version, if they are ran in extendr-ci. We have already gone through this, but I think some things fell through the cracks.

Issue #387, opened 2024-09-17
use_crate fails when no features specified


JosiahParry:
This change broke the function. Looks like we might be missing a test for this. I installed the new version 

```
rextendr::use_crate("lazy-static")
Running cargo add lazy-static '--features ' '--optional false'
Error:
! ! System command 'cargo' failed
```

This is because `"--features "` is included in the vector that is passed to processx. `--features`

_Originally posted by @JosiahParry in https://github.com/extendr/rextendr/pull/361#discussion_r1764103981_

CGMossa:
Fixes by #388 (thanks @kbvernon for both the PR, and for notifying me that this issue can now be closed)

PR #388, opened 2024-09-18
check for empty string in use_crate()

kbvernon:
To address https://github.com/extendr/rextendr/issues/387, this PR adds a check for empty strings when running `purrr::discard()` on the `cargo_add_opts` list.


CGMossa:
How do we know that this works? Is it possible to add a test? Otherwise I'd just merge.

kbvernon:
yes, this function needs a test. i'll look into adding one tomorrow!

kbvernon:
Pasting flags and values together into single character scalars is the wrong way to pass additional arguments to `processx::run()`, so I opted to concatenate them into a character vector. This has the advantage of making `purrr::discard()` unnecessary, as concatenating drops NULL values.

tests/testthat/test-use_crate.R
```diff
@@ -0,0 +1,28 @@
+test_that("use_crate() adds dependency to package or workspace", {
```
Ilia-Kosenkov, line 1: Could you please add a case for unhappy path? E.g. both version and git?

kbvernon: i also added the "rextendr_error" class to the `cli_abort()` call

JosiahParry:
@kbvernon one last test! can you test the optional feature? The nifty codecov tool shows that it is missing a check :)

JosiahParry:
100% code cov? You're crazy for that.

JosiahParry (approved):

PR #393, opened 2024-10-27
Add `--print=native-static-libs` when compiling 

JosiahParry:
Per suggestion of Yutanni, we should add `--print=native-static-libs` which gives us insight into which linker flags need to be visible when building our crate.

If a user encounters an error building, they can check the output.

### DIFF
--- a/.github/workflows/test_pkg_gen.yaml
+++ b/.github/workflows/test_pkg_gen.yaml
@@ -85,17 +85,6 @@ jobs:
             file.path("tests", "testthat", "test-hello.R")
           )
 
-          # test NOT_CRAN envvar
-          brio::write_lines(
-            c(
-              "if [ \"$NOT_CRAN\" != \"true\" ]; then",
-              "  exit 1",
-              "fi"
-            ),
-            file.path("configure")
-          )
-          Sys.chmod("configure", "0755")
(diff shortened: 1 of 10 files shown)


JosiahParry:
Note that the snapshot tests are failing because between the version of {usethis} that I have `2.2.3` and the version being tested on main `3.0.0.` there is a period placed at the end of the `use_build_ignore()` message: 

```diff
-   v Adding "^src/rust/vendor$" to '.Rbuildignore'.
+   v Adding '^src/rust/vendor$' to '.Rbuildignore'
-   v Adding "src/rust/vendor" to '.gitignore'.
+   v Adding 'src/rust/vendor' to '.gitignore'
```

This has taken a while to debug. Perhaps we can reduce what the snapshot tests are actually testing.

JosiahParry:
This PR also fixes the R-devel errors. It does this by adding the `configure` and `tools/msrv.R` scripts from `use_cran_defaults()` into the standard `use_extendr()`.

CGMossa (approved):

CGMossa:
We need to make this flag additive to if it was set by the user to contain more options though.

And now that apparently all package tests are cran tests... I don't know what we do about that.

PR #397, opened 2024-11-18
refactor cargo-processx functions

kbvernon:
This PR tries to bring consistency to how functions call `processx::run()`. The main goal is to unburden this package by letting `processx` do the heavy lifting wherever possible. Changes include:
- setting `error_on_status = TRUE` everywhere
- adding an `echo` parameter and passing it to `echo_cmd = echo` (print cargo command to console) and `echo = echo` (print stdout to console). The default is `TRUE` to ensure that maximum information is printed to the console. 
- setting `env = get_cargo_envvars()`. I'm not sure this one is strictly necessary. I believe the function was originally written to handle issues with rtools and windows, but functions that don't use this still pass CI on all OS.
- standardize logic for defining `wd`. This is now done with a single call to `rprojroot::find_package_root_file()`.
- using `jsonlite::parse_json(simplifyDataFrame = TRUE)` whenever cargo command returns json formatted strings in stdout. If we can always assume data.frames rather than lists, I think this will simplify future maintenance.

At the same time, I took the liberty of adding standalone type checks, reformatting to ensure scripts past lintr tests, removing magrittr pipes, and replacing redundant code. 

Right now, there are seven files with calls to `processx::run()`, but I only updated four of them: 
- [x] clean.R 
- [ ] cran-compliance.R
- [x] license_note.R
- [x] read_cargo_metadata.R
- [ ] source.R
- [x] use_crate.R
- [ ] utils.R

I'm ignoring cran-compliance.R because I think Josiah's recent work will make that code obsolete. I think the myriad issues in `source.R` are substantial enough that they call for their own PR. And the call in `utils.R` is only really there to check for third party cargo tools, so maybe not worth messing with.


tests/testthat/test-license_note.R
```diff
@@ -7,6 +7,4 @@ test_that("LICENSE.note is generated properly", {
   write_license_note()
 
   expect_snapshot(cat_file("LICENSE.note"))
-  expect_rextendr_error(write_license_note(), NA)
-  expect_rextendr_error(write_license_note(force = FALSE), "LICENSE.note already exists.")
```
JosiahParry, line 11: Why were these tests removed?

R/license_note.R
```diff
@@ -1,97 +1,120 @@
...
+
+  outfile <- rprojroot::find_package_root_file(
+    "LICENSE.note",
+    path = path
+  )
+
+  metadata <- read_cargo_metadata(
+    path = path,
+    dependencies = TRUE
+  )
```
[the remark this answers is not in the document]
JosiahParry: If the error case is handled here then the codecov below should be fine. but essentialy we're not checking that the Cargo.toml file exists. we should do that and error accordingly

R/license_note.R
```diff
@@ -1,97 +1,120 @@
...
-  if (!isTRUE(force) && file.exists(outfile)) {
-    cli::cli_abort(
-      c(
-        "LICENSE.note already exists.",
-        "If you want to regenerate LICENSE.note, set `force = TRUE` to {.fun write_license_note}."
-      ),
-      class = "rextendr_error"
-    )
+  # exclude current package from LICENSE.note
+  current_package <- metadata[["resolve"]][["root"]]
```
JosiahParry, line 45: check if null and error if null

tests/testthat/test-license_note.R
kbvernon: A couple of things:
- [`write_file()`](https://github.com/extendr/rextendr/blob/main/R/write_file.R) has an `overwrite` parameter. I updated the parameter name in `write_license_note()` from `force` to `overwrite` to make it consistent.
- `write_file()` also has its own check to see if the file exists, but it only emits a warning and returns NULL rather than an error. I figured this check only really needs to be handled in one place, so I removed it from `write_license_note()`.  

That means the test can't pass, so I was thinking I could either re-write it or update `write_file()` to emit an error. But for now, I just removed those tests.

Does this count as mission creep?

R/license_note.R
```diff
@@ -1,97 +1,120 @@
...
+  packages[["licenses"]] <- replace_na(
+    packages[["repository"]],
+    "not provided"
+  )
+
+  prep_authors <- function(authors, package) {
+    authors <- ifelse(
+      is.na(authors),
+      paste0(package, " authors"),
+      authors
```
JosiahParry, line 74: can we move out of inline?

R/license_note.R
```diff
@@ -1,97 +1,120 @@
...
+    authors <- stringi::stri_replace_all_regex(authors, r"(\ <.+?>)", "")
+
+    paste0(authors, collapse = ", ")
   }
 
+  packages[["authors"]] <- unlist(Map(
+    prep_authors,
+    packages[["authors"]],
+    packages[["name"]]
+  ))
```
JosiahParry, line 86: code cov wants us to check that `packages[["authors"]]` and `packages[["name"]]` is not null and the correct type. We can add some top level checks after `packages` is created with errors being thrown in the appropriate condition.

R/license_note.R
```diff
@@ -1,97 +1,120 @@
...
   separator <- "-------------------------------------------------------------"
 
-  note_header <- paste0(
-    "The binary compiled from the source code of this package contains the following Rust crates:\n",
-    "\n",
-    "\n",
-    separator
+  note_header <- glue::glue(
+    "
+    The binary compiled from the source code of this package \\
```
JosiahParry, line 92: why did we add `\\` here ?

R/license_note.R
```diff
@@ -1,97 +1,120 @@
...
-      }
-    )
+  note_body <- glue::glue_data(
+    packages,
+    "
+
+    Name:        {name}
+    Repository:  {repository}
+    Authors:     {authors}
+    License:     {license}
```
JosiahParry, line 107: can we use paste or something similar here that isn't `glue::glue_data()` this is going to make it harder to remove glue at a later point.

R/use_crate.R
```diff
@@ -43,14 +45,15 @@ use_crate <- function(
...
-  check_string(path)
+    path = ".",
+    echo = TRUE) {
+  check_string(crate, class = "rextendr_error")
+  check_character(features, allow_null = TRUE, class = "rextendr_error")
+  check_string(git, allow_null = TRUE, class = "rextendr_error")
+  check_string(version, allow_null = TRUE, class = "rextendr_error")
+  check_bool(optional, class = "rextendr_error")
+  check_string(path, class = "rextendr_error")
+  check_bool(echo, class = "rextendr_error")
```
Ilia-Kosenkov, line 56: Do we anywhere use different `class`? Why not embed it into functions by default? Or completely?

R/read_cargo_metadata.R
```diff
@@ -25,22 +29,44 @@
...
+      "--no-deps"
+    },
+    if (tty_has_colors()) {
+      "--color=always"
+    } else {
+      "--color=never"
+    }
   )
 
   out <- processx::run(
```
Ilia-Kosenkov, line 58: Why not extract calls to `processx::run()` together with error handling and input validation and reuse it everywhere?

R/license_note.R
kbvernon: it's `glue` for "ignore line break." i broke up the line to keep it in the 80 character limit.

R/clean.R
```diff
@@ -28,52 +39,34 @@ clean <- function(path = ".") {
...
-    glue("--manifest-path={toml_path}"),
-    glue("--target-dir={target_dir}"),
+    glue::glue("--manifest-path={manifest_path}"),
+    glue::glue("--target-dir={target_dir}"),
     if (tty_has_colors()) {
       "--color=always"
     } else {
       "--color=never"
-    },
-    "--quiet"
```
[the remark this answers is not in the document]
kbvernon: seems better to handle that with `echo_cmd` and `echo` in `processx::run()`, right?

R/clean.R
```diff
@@ -28,52 +39,34 @@ clean <- function(path = ".") {
...
-    error_on_status = FALSE,
-    env = cargo_envvars
+    error_on_status = TRUE,
+    wd = rust_folder,
+    echo_cmd = echo,
+    echo = echo,
+    env = get_cargo_envvars()
   )
 
-  if (!isTRUE(exec_result$status == 0)) {
```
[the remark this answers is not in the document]
kbvernon: i think `processx::run(error_on_status = TRUE)` does this, and it passes on the cargo error message. this is one of the main changes i wanted to make more consistent across our cargo utilities.

R/license_note.R
```diff
@@ -1,97 +1,120 @@
...
+    "
+  )
 
   write_file(
     text = c(note_header, note_body),
     path = outfile,
     search_root_from = path,
-    quiet = quiet
+    quiet = quiet,
+    overwrite = overwrite
```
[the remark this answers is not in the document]
kbvernon: we were checking for the existence of `LICENSE.note` twice, once in `write_license_note()` and once in `write_file()`. it seemed to make sense to just let `write_file()` do that, but maybe not?

R/read_cargo_metadata.R
```diff
@@ -25,22 +29,44 @@
...
+    error_on_status = TRUE,
+    wd = rust_folder,
+    echo_cmd = echo,
+    echo = echo,
+    env = get_cargo_envvars()
   )
 
-  jsonlite::fromJSON(out[["stdout"]])
+  jsonlite::parse_json(
+    out[["stdout"]],
```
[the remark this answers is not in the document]
kbvernon: yeah, this is not something i had considered before. josiah has also requested that i add checks any time `[[` is used, so i will do that asap.

R/read_cargo_metadata.R
kbvernon: i've added a draft of a `run_cargo()` function that handles all `processx::run()` logic

R/use_crate.R
JosiahParry: We don't. I think we can address this in a followup PR? One of the  challenges i foresee is that class is set via `...` here. The bulk of these checks have moved into the `run_cargo()` which reduces amount of the `class = "rextendr_error"` instances

R/read_cargo_metadata.R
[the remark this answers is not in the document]
JosiahParry: handled using `rlang::try_fetch()`

tests/testthat/test-license_note.R
JosiahParry: yes it definitely does! Let's create a new issue and add it to the 0.4.0 milestone

JosiahParry:
codcov is being a bit extreme here. We can address the coverage later on. It is actually penalizing this PR for adding more checks funnily enough. 

Good work @kbvernon

JosiahParry (approved):

PR #399, opened 2024-11-20
internal utilities to find rust crate and cargo manifest

kbvernon:
This PR adds internal utility functions that find R package root, build path from there to crate directory or Cargo.toml, and check that directory or file exists. Both return paths. These can be used to consolidate code and provide stronger checks. 

Note: this is to simplify efforts to refactor functions to execute Cargo subcommands using `processx::run()` in PR https://github.com/extendr/rextendr/pull/397.

### DIFF
--- a/R/find_extendr.R
+++ b/R/find_extendr.R
@@ -44,6 +44,8 @@ find_extendr_crate <- function(
 #' @examples
 #' \dontrun{
 #' find_extendr_manifest()
+#' @keywords internal
+#' @noRd
 #' }
 find_extendr_manifest <- function(
     path = ".",
--- a/R/find_extendr.R
+++ b/R/find_extendr.R
@@ -9,6 +9,8 @@
 #' @examples
 #' \dontrun{
 #' find_extendr_crate()
+#' @keywords internal
+#' @noRd
 #' }
 find_extendr_crate <- function(
     path = ".",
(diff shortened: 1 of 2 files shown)


R/find_extendr.R
```diff
@@ -0,0 +1,65 @@
...
+#' @param path character scalar, the R package directory
+#' @param error_call call scalar, from rlang docs: "the defused call with which
+#' the function running in the frame was invoked"
+#'
+#' @return character scalar, path to Rust crate
+#'
+#' @examples
+#' \dontrun{
+#' find_extendr_crate()
+#' }
```
JosiahParry, line 12:
```suggestion
#' @keywords internal
#' @noRd
#' }
```

R/find_extendr.R
```diff
@@ -0,0 +1,65 @@
...
+#' @param error_call call scalar, from rlang docs: "the defused call with which
+#' the function running in the frame was invoked"
+#'
+#' @return character scalar, path to Cargo manifest
+#'
+#' @examples
+#' \dontrun{
+#' find_extendr_manifest()
+#' }
+find_extendr_manifest <- function(
```
JosiahParry, line 46:
```suggestion
#' @keywords internal
#' @noRd
#' }
find_extendr_manifest <- function(
```

JosiahParry (changes requested):
Looks good! we need to just ensure that the documentation doesn't actually get rendered. That will lead to a rejected package in CRAN

JosiahParry (approved):

PR #405, opened 2024-11-22
chore: increase code coverage

JosiahParry:
This PR seeks to improve the overall code coverage of {rextendr} without using snapshot tests. 

This will hopefully help us identify any regressions in the future.

### DIFF
--- a/R/cran-compliance.R
+++ b/R/cran-compliance.R
@@ -25,12 +25,12 @@ vendor_pkgs <- function(path = ".", quiet = FALSE, overwrite = NULL) {
   local_quiet_cli(quiet)
 
   # get path to rust folder
-  src_dir <- rprojroot::find_package_root_file(path, "src/rust")
+  src_dir <- rprojroot::find_package_root_file("src", "rust", path = path)
 
   # if `src/rust` does not exist error
   if (!dir.exists(src_dir)) {
     cli::cli_abort(
-      c("{.path src/rust} cannot be found", "i" = "Did you run {.fn use_extendr}?"),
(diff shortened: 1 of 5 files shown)


R/cran-compliance.R
```diff
@@ -25,12 +25,12 @@ vendor_pkgs <- function(path = ".", quiet = FALSE, overwrite = NULL) {
   local_quiet_cli(quiet)
 
   # get path to rust folder
-  src_dir <- rprojroot::find_package_root_file(path, "src/rust")
+  src_dir <- rprojroot::find_package_root_file("src", "rust", path = path)
```
kbvernon, line 28: This is `find_extendr_crate()`. The error message here is better though, maybe we consider updating both `find_extendr_*` functions to use it? Also, should we wrap all `find_package_root_file()` calls in `rlang::try_fetch()`?

JosiahParry: This call doesn't cause an error if the path doesn't exist which is why the below check exists. 

It may make sense to update `find_extendr_crate()` later on, but I think it is out of scope of this PR. 

As an aside I think the `path` is not set appropriately elsewhere. See how here it was modified to setting the `path` argument?

JosiahParry: It was passed through dots but it should not have see: 

```r
args(rprojroot::find_package_root_file)
#> function (..., path = ".") 
#> NULL
```

That sounds good, but I still think it is out of scope of this PR since this PR is focused on increasing test coverage. This change was made to improve test coverage. The code below it was never reached because of the misordered arguments

Ilia-Kosenkov (approved):

PR #408, opened 2024-11-26
use standalone purrr

JosiahParry:


JosiahParry:
This PR removes {purrr} as a dependency by using the standalone purrr from https://github.com/r-lib/rlang/blob/main/R/standalone-purrr.R

This doesn't have full coverage of purrr functionality but I believe I've handled all the edge cases.

README.Rmd
```diff
@@ -133,10 +133,10 @@ y <- c(1, 2, 3, 4, 5)
...
   Name = c("x", "y"),
   Data = list(x, y),
-  Types = purrr::map_chr(Data, typeof),
-  Sum = purrr::map(Data, get_sum),
-  SumRaw = purrr::flatten_dbl(Sum),
-  ResultType = purrr::map_chr(Sum, typeof)
+  Types = map_chr(Data, typeof),
+  Sum = map(Data, get_sum),
+  SumRaw = flatten_dbl(Sum),
+  ResultType = map_chr(Sum, typeof)
```
Ilia-Kosenkov, line 139: This is an example, if you do not export shims, the caller will not have access to this functions, will they?

Ilia-Kosenkov:
Dropping a dependency is worth mentioning on the `NEWS.md`.

R/toml_serialization.R
```diff
@@ -122,7 +123,8 @@ format_toml.data.frame <- function(x,
                                    .tbl_name,
                                    .top_level = FALSE) {
   rows <- nrow(x)
-  header <- glue("[[{.tbl_name}]]")
+  # header <- glue("[[{.tbl_name}]]")
```
Ilia-Kosenkov, line 126: If you aim at throwing away `glue`, why comment and not remove?

JosiahParry: ah good catch, i was trying to find an error that was coming from glue and I wanted to see if it was from here. This can be either removed or reverted to glue. Only focus in this PR is for purrr at the moment.

Ilia-Kosenkov (approved):

JosiahParry:
<3 thank you!

Issue #375, opened 2024-08-06
set minimum R version and minimum Rust version in `use_extendr()`


etiennebacher:
Related to https://github.com/etiennebacher/astgrepr/issues/28

Compiling `astgrepr` from source fails on R 4.0.3 with this error:

```
error[E0425]: cannot find value `S4SXP` in this scope
   --> /tmp/RtmpRgdx7u/R.INSTALL1ae525e0ee11/astgrepr/src/vendor/extendr-api/src/lib.rs:574:15
    |
574 |         S4 => S4SXP,
    |               ^^^^^ not found in this scope

For more information about this error, try `rustc --explain E0425`.
error: could not compile `extendr-api` (lib) due to previous error
warning: build failed, waiting for other jobs to finish...
make: *** [Makevars:19: rust/target/release/libastgrepr.a] Error 101
```

I suppose this is because `extendr` takes advantage of new features in R, so compiling from source won't work for older versions of R. But does this mean users on R 4.0 won't be able to install the binary as well?

More generally, does using `extendr` affect the minimal R version supported? If so, this should be documented somewhere but I didn't find anything on this.

CGMossa:
So I think we have a feature gate for this, so it should work for older versions of R. In theory. Something could have slipped up.

I believe we only don't support the R version that still had 32-bit support, because that's too much for us to handle. But that's an unofficial policy that I'm just speaking of here.

There could be a bug, I'm unsure :( I don't have time to investigate unfortunately. But if you do try to investigate it, I'm sure you'd find a bug somewhere.. I have feature flags in extendr for this.

etiennebacher:
I also don't have time to search for this, especially since I'm not familiar at all with the codebase, but I think a message when initiating `rextendr` (or even automatically update the `DESCRIPTION` with `Depends: R (>= x.y.z)`) would be welcome.

etiennebacher:
Related: https://github.com/extendr/extendr/issues/664

Ilia-Kosenkov:
You might be able to get away by referencing older versions of `extendr`. `{rextendr}` itself depends on R >= 4.0, and should not use features from later versions. https://github.com/extendr/rextendr/blob/c4f5b0cea27f144edc2544bbfed1f35f505c3aa8/DESCRIPTION#L40

We provide support for current, oldrel and devel versions of R (current, previous and future minor versions).

[...]

PR #414, opened 2025-02-20
fix: Remove unnecessary makevars.ucrt

### LINKED ISSUE #332: Makevars.ucrt is no longer needed
unless you want to support R <= 4.1.

albersonmiranda:
Removes `makevars.ucrt` templates and updates tests and snaps. Closes #332 .

### DIFF
--- a/.lintr
+++ b/.lintr
@@ -4,3 +4,6 @@ linters: linters_with_defaults(
     commented_code_linter = NULL
     )
 encoding: "UTF-8"
+exclusions: list(
+      "R/import-standalone-obj-type.R"
+    )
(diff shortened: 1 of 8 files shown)


JosiahParry:
Thank you @albersonmiranda ! The changes to libR-sys actually require that the minimum supported version of R is 4.2. 

I think we should set the minimum supported version of R to 4.2. see https://github.com/extendr/rextendr/issues/375

JosiahParry:
@albersonmiranda can you add a note to the changelog? I think we can skip the lintr failure right now and add exclusions for the standalone files using the .lintr file https://lintr.r-lib.org/articles/lintr.html at a later point

JosiahParry (approved):

PR #429, opened 2025-04-13
Remove `pandoc` / `quarto` from CI runs

CGMossa:
Evidently, `pandoc` is required for the vignette checks, so it is staying for those.

Tangential discovery: Quarto is not installed unless it detects files that needs it (`qmd`), however `pandoc` is installed by default always.


Ilia-Kosenkov (approved):

CGMossa:
Did anyone check if this PR actually improves ci time?

Issue #428, opened 2025-04-13
tests: trigger rextendr CI when extendr PR is merged


JosiahParry:
We removed interlocking testing from extendr (see https://github.com/extendr/extendr/pull/888/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88). However, we still need to ensure that rextendr tests stay up to date with extendr changes. 

The best way for us to do this is to trigger tests in rextendr when we make a change to extendr. 

We can probably handle it with [repository_dispatch](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#repository_dispatch) workflow events. 

See related: https://github.com/peter-evans/repository-dispatch

PR #430, opened 2025-04-13
Add workflow to run checks when pr merged event received from extendr

brancengregory:
Currently only triggers R-CMD-check.yml workflow. Do you want others added @JosiahParry?

### DIFF
--- a/.github/workflows/extendr_update.yaml
+++ b/.github/workflows/extendr_update.yaml
@@ -1,0 +1,8 @@
+name: Run Tests on extendr Update
+on:
+	repository_dispatch:
+		types: [extendr-pr-merged]
+
+jobs:
+	run_rextendr_tests:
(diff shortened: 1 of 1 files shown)


brancengregory:
Closes https://github.com/extendr/rextendr/issues/428

JosiahParry:
I think this is great!

.github/workflows/extendr_update.yaml
```diff
@@ -0,0 +1,8 @@
+name: Run Tests on extendr Update
+on:
+	repository_dispatch:
+		types: [extendr-pr-merged]
+
+jobs:
+	run_rextendr_tests:
+		uses: ./.github/workflows/R-CMD-check.yaml
```
JosiahParry, line 8: i think the indentation may be off here. we'll see what happens

JosiahParry (approved):

PR #434, opened 2025-04-25
Use vendored dependencies if present

albersonmiranda:
Simplify the logic for unzipping the vendor.tar.xz file by removing the NOT_CRAN condition, ensuring that the vendored dependencies are utilized when available.

This change helps avoid MSRV breakages caused by dependency updates, which may occur even when there are no changes in `cargo.toml`.


CGMossa:
I can't review the validity of this. I just want to say, that breaking MSRV is not protected by minor version changes in the semver, and what happens now, is an adherence to the rust-version specificied (if it is) in the `Cargo.toml`.  See https://doc.rust-lang.org/cargo/reference/resolver.html?highlight=rust-version#rust-version

CGMossa:
Just so you know, the patched rtools version has to match the versions. See my commit. This alteration has to be reflected in the rextendr code for rtools selection.

Ilia-Kosenkov (approved):

PR #433, opened 2025-04-20
draft: remove tibble dependency

JosiahParry:

PR #435, opened 2025-04-26
Feature/remove tibble

Ilia-Kosenkov:
Overtaking #433 to slightly speed up things

### DIFF
--- a/tests/testthat/_snaps/cran-compliance.md
+++ b/tests/testthat/_snaps/cran-compliance.md
@@ -14,17 +14,15 @@
     Code
       package_versions
     Output
-      # A tibble: 10 x 2
-         crate          version
-         <chr>          <chr>  
-       1 build-print    *.*.*
-       2 extendr-api    *.*.*
-       3 extendr-ffi    *.*.*
-       4 extendr-macros *.*.*
-       5 once_cell      *.*.*
-       6 paste          *.*.*
-       7 proc-macro2    *.*.*
-       8 quote          *.*.*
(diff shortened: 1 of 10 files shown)


tests/testthat/_snaps/cran-compliance.md
```diff
@@ -14,17 +14,15 @@
...
+      1     build-print   *.*.*
+      2     extendr-api   *.*.*
+      3     extendr-ffi   *.*.*
+      4  extendr-macros   *.*.*
+      5       once_cell  *.*.*
+      6           paste  *.*.*
+      7     proc-macro2  *.*.*
+      8           quote  *.*.*
+      9             syn *.*.*
+      10  unicode-ident  *.*.*
```
JosiahParry, line 27: Given that we don't actually test the versions here. I'm of the mind we should extract the crate column and check against a known vector. This snapshot in particular has been a bit...finicky. Though I suppose it is out of scope!

JosiahParry (approved):

tests/testthat/_snaps/cran-compliance.md
Ilia-Kosenkov: I haven't noticed any flakyness, but I might be wrong. If it will constantly cause issues, yup, let's replace it

CGMossa:
Specifying the rtools version is unnecessary. That was my fault.

CGMossa:
Specifying the rtools version in the work flow file is not necessary

PR #444, opened 2025-05-02
Various changes to path handling 

CGMossa:
Due to failings of the CRAN submission, I've hacked together these slight changes.

I've removed `export`s, as I suspect, something funny might be happening with the `PATH` definitions.

Then, I've removed nested `withr`/`processx` calls in `vendor_pkgs`. Note, that `vendor_pkgs` is the culprit in all ERROR for rextendr 0.4.0.

I've changed the file-separator in `configure`/`configure.win` related scripts. I am still not sure that this does anything, but nevertheless it is more correct. 

I tried desperately to reproduce the error locally and on CI, but was not able to.


JosiahParry:
Hm..isn't the cargo home necessary for CRAN because they do not permit us to modify anything globally which cargo will do? IIUC the point of the CARGO_HOME env var is to ensure that the .cargo folder is relative to the R package and not the users home

Ilia-Kosenkov:
So what is the verdict here?

R/cran-compliance.R
```diff
@@ -39,17 +39,16 @@ vendor_pkgs <- function(path = ".", quiet = FALSE, overwrite = NULL) {
...
-        stderr_line_callback = stderr_line_callback
-      )
-    })
+    update_res <- processx::run(
+      wd = src_dir,
+      "cargo",
+      c(
+        "generate-lockfile",
+        "--manifest-path",
+        file.path(src_dir, "Cargo.toml")
```
Ilia-Kosenkov, line 48: This looks strange to me TBH, how does path relative to `src_dir` interplay with setting `wd` to `src_dir`?

CGMossa: I agree. As you can see, I modified it only gently.

albersonmiranda:
Since we can't reproduce the error and therefore can't confirm it's fixed, I suggest we close this PR and skip all tests that call `vendor_pkgs()` on CRAN. This is mainly because `vendor_pkgs()` triggers downloads from crates.io, which is not allowed on CRAN. These tests are destined to fail!

CGMossa:
> Since we can't reproduce the error and therefore can't confirm it's fixed, I suggest we close this PR and skip all tests that call `vendor_pkgs()` on CRAN. This is mainly because `vendor_pkgs()` triggers downloads from crates.io, which is not allowed on CRAN. These tests are destined to fail!

I agree. Can you make a PR that does that?

Also, this PR has some "corrections" that I think we may agree on regardless of bug-fixes. For instance, the path separator, and how environment variables are ought to be specified. The nested `withr::with_dir` and `processx` might also be a bit confusion.. Feel free to lift these changes to beef up your PR.

PR #475, opened 2025-09-17
feat: add configure arguments to enable features or profiles during installation

Yunuuuu:
This PR introduces two new configure arguments, `--with-features` and `--with-profile`, which allow enabling features or profiles during installation.

It also respects the environment variables `{pkg_name}_FEATURES` and `{pkg_name}_PROFILE`.

### DIFF
--- a/R/use_extendr.R
+++ b/R/use_extendr.R
@@ -201,7 +201,8 @@ use_extendr <- function(
     "config.R",
     save_as = file.path("tools", "config.R"),
     quiet = quiet,
-    overwrite = overwrite
+    overwrite = overwrite,
+    data = list(pkg_name = pkg_name)
   )
 
   # add configure and configure.win templates
--- a/inst/templates/Makevars.in
+++ b/inst/templates/Makevars.in
@@ -30,7 +30,7 @@ $(STATLIB):
 
 	export CARGO_HOME=$(CARGOTMP) && \
 	export PATH="$(PATH):$(HOME)/.cargo/bin" && \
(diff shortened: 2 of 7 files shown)


Yunuuuu:
Thanks for the comment. I need the `features` flag to control the usage of test and benchmark functionality.

If configuration arguments are not supported, at the very least there should be a `features` environment variable like `profile` so users can explicitly select which features to enable.

For example, the following function should only be available when the bench feature is enabled:

```rust
#[extendr]
#[cfg(feature = "bench")]
fn pprof_kractor_koutput(
    pprof_file: &str,
) -> std::result::Result<(), String> {
    let guard = pprof::ProfilerGuardBuilder::default()
        .frequency(2000)
        .build()
        .map_err(|e| format!("cannot create profile guard {:?}", e))?;
    // ...; actual code
    if let Ok(report) = guard.report().build() {
        let file = std::fs::File::create(pprof_file).map_err(|e| {
            format!("Failed to create file {}: {}", pprof_file, e)
        })?;
        let mut options = pprof::flamegraph::Options::default();
        options.image_width = Some(2500);
        report
            .flamegraph_with_options(file, &mut options)
            .map_err(|e| {
                format!("Failed to write flamegraph to {}: {}", pprof_file, e)
            })?;
    };
    out
}
```

Regarding the test issue: it cannot be fixed within the unit test file. The reason is that `use_extendr()` checks `is_vscode() || is_positron()` to decide whether `use_vscode()` should be applied. If tests are executed from VS Code, a failure will always occur because the snapshot is generated without VS Code support.

Yunuuuu:
I want to clarify that this does not depend on the first or second argument, but specifically on the `--with-features` and `--with-profile` flags passed through `install.packages(configure.args = "--with-features=...")`. In the current implementation (without this PR), `configure.args` is completely ignored.

Yunuuuu:
`configure.args` is also described in the R Extensions manual as a way to control the build process, which can be applied when using `R CMD INSTALL`:
<https://rstudio.github.io/r-manuals/r-exts/Creating-R-packages.html#configure-example>

image

PR #479, opened 2025-10-01
Better vendor experience

albersonmiranda:
This PR:

1. Enables the user to opt in to cleaning the vendor directory after creating the tarball: `vendor_pkgs(clean = TRUE)`. Defaults to `FALSE`.
2. Updates `Makevars` so it looks for `src/vendor/` first. If not present, it will then unzip the tarball (useful for projects with heavy dependencies).


JosiahParry (approved):
THank you!

PR #520, opened 2026-06-16
update hello-world example

kbvernon:
This pr updates the hello-world example in the lib.rs template to show a round trip of data from R to Rust then back to R, i.e. `hello_world()` becomes `hello("world")`. This will mainly be useful for examples in the documentation.


JosiahParry (approved):
@kbvernon thank you!

